Issue #108, opened 2015-01-14
Speed up DS18B20


gfwilliams:
The `writeSpad` function uses 'arguments', which is quite slow as every mention of it causes the array to be recreated. It'd be better explicitly labelling each write.

gfwilliams:
Fixed now

PR #109, opened 2015-01-14
Address issue #108

bartmichu:
> It'd be better explicitly labelling each write.

Hi, you mean like this?


gfwilliams:
That's perfect! Thanks for the PR!

PR #115, opened 2015-01-28
Append links to eeprom example, add ebay buying links to device docs

SpenceKonde:
Recently some one who surely had no conflict of interest at all, went through and added buying links to all the device docs, linking to one specific vendor, digitalmeans, whose prices are exorbitant (except in the case where the site didn't actually sell the device). So everyone who read the pages was getting links suggesting that these parts cost 5-10x what they're available for from china-via-ebay. 

In each place where there weren't other vendors, I added an ebay link (to a search for the part number within the  business and industrial -> electronic and test equipment section, so readers get to see more than one source, at least one of which has the product at reasonable prices.

### DIFF
--- a/devices/ADXL335.md
+++ b/devices/ADXL335.md
@@ -32,4 +32,5 @@ Buying
 
 ADXL335 modules can be purchased from many places:
 
+* [eBay](http://www.ebay.com/sch/i.html?_nkw=ADXL335&_sacat=92074)
 * [digitalmeans.co.uk](https://digitalmeans.co.uk/shop/index.php?route=product/search&tag=adxl335)
--- a/devices/ADXL345.md
+++ b/devices/ADXL345.md
@@ -174,5 +174,5 @@ Buying
 -----
 
 ADXL345 modules can be purchased from many places:
-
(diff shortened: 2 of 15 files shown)


gfwilliams:
Great - thanks!

I'm trying to build up a list of suppliers, in the hope that at some point I'll be able to have a link on each project that says 'you can buy what you need here'. As it happens, digitalmeans was willing to go through and add links to their stuff. Even though they might be more expensive, at least there's a place for people to go if they're be willing to look elsewhere for a better deal.  

If only more companies were willing to add links too :)

Was there a reason digitalmeans got taken out for the BMP180 (BMP085.md) and SmartNixie? I guess their nixie might not be compatible, but the BMP180 looks ok?

SpenceKonde:
There were two links. I thought they were duplicates, so I deleted one of them. I now see that they linked to both 180 and 085, I restored those links, but formatted them to only take up a single line. There. 

And the link on smart nixie was removed because they don't sell the SmartNixie device, and the devices the link pointed to were unrelated nixie-like tubes.

gfwilliams:
Great,  thanks!

PR #117, opened 2015-02-01
Add support for UserName/Password Authentication to MQTT module

rockymanobi:
Hi Gordon.

I add support for UserName / Password Authentication to MQTT module written by `@ltj`. Thanks. 

The changes are as follow.
* Make `MQTT#constructor' accept the options below.
  * username
  * password
  * protocol_level
  * protocol_name
* Add information about options of `connect` method to `MQTT.md`.

### Test

I tested the code with brokers below.

*  [RabbitMQ-MQTT Adaptor](http://www.rabbitmq.com/mqtt.html)
*  mqtt.js based broker ( [source code](https://github.com/rockymanobi/espruino-mqtt-sample/tree/master) )

### DIFF
--- a/devices/MQTT.js
+++ b/devices/MQTT.js
@@ -6,8 +6,8 @@ Simple MQTT protocol wrapper for Espruino sockets.
 /** 'private' costants */
 var C = {
   PACKET_ID      : 1,     // Bad...fixed packet id
-  PROTOCOL_LEVEL : "\x04" // MQTT protocol level
-}
+  PROTOCOL_LEVEL : 4 // MQTT protocol level
+};
 
 /** Control packet types */
 var TYPE = {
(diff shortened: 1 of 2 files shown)


gfwilliams:
Looks awesome - thanks!

ltj:
Great @rockymanobi . I was about to add username/passwd support myself when I saw you saved me the effort :)

One question though - does any of the brokers you mention support other protocol names than "MQTT"? According to the OASIS standard they may, but then it's no longer MQTT compliant. Just curious to know if this is really a needed feature. Thanks.

If you are interested @rockymanobi, we could try and collaborate on making this module even more useful.

Cheers,
Lars

rockymanobi:
@ltj 

Thanks for reply, and sorry for my late reply.

> If you are interested @rockymanobi, we could try and collaborate on making this module even more useful.

Yes, I'm really interested in. Actually, I was about to ask the same thing. It will be great if the Espruino will support MQTT completely, and we should try and collaborate.

> One question though - does any of the brokers you mention support other protocol names than "MQTT"? According to the OASIS standard they may, but then it's no longer MQTT compliant. Just curious to know if this is really a needed feature. Thanks.

The protocol names I mention to support are "MQIsdp"  and "SCADA".  They are both old names for MQTT.  And there are some brokers which sitll expect "MQIsdp" or "SCADA" as protocol_name, such as RabbitMQ-MQTT-plugin. 

Thanks!

PR #119, opened 2015-02-21
Minor changes to MQTT module

ltj:
Small fixes to MQTT module after @rockymanobi 's excellent addition of username/password support:
- Password flag only set if username flag is asserted (per MQTT OASIS standard)
- Likewise password is only added to the payload if username is also present
- standard object creation now in stead of my weird Object.create approach
- minor typos etc fixed


gfwilliams:
Great - thanks!

PR #123, opened 2015-03-28
Added ESP8266 code and viewport tags

the1laz:
Added to tutorial for Wifi remote console.  Alternative code for running on  ESP8266 added at the end of the example. Viewport tags added to each piece of code to make the page load correctly on mobile devices as well.


gfwilliams:
Great, thanks! Was this done with the totally standard ESP8266 module in Espruino?

the1laz:
Yep. I used different pins to run it on my board as I've got a Nucleo, but I used the pins from the ESP8266 tutorial in this tutorial to keep things the same.

PR #132, opened 2015-04-22
Stops Hanging Code if INIT failed

willoftw:
Swift change to stop hanging if init failed (eg bad ID). and returns Null, so you don't get an object nocking around that cant do anything with.


gfwilliams:
Great - thanks!

PR #134, opened 2015-04-26
Brain.js module for interfacing the NeuroSky ThinkGear ASIC module (TGAM)

pastaclub:
Hi, I have developed an Espruino module for interfacing the TGAM, an EEG module which allows people to measure brainwaves. The TGAM has a serial interface and can easily be hooked up to an Espruino board using only a UART RX pin, 3.3V power and GND.

I thought about how to name this module. The hardware is referred to by various names, the most common one being TGAM. However, there is a library for the Arduino which interfaces this device and that library is called "Brain", so I figured that it would be best to call this one "Brain" as well. Note that this is not ported from the Arduino. It's an original development in JavaScript following the protocol description of the manufacturer.

### DIFF
--- a/devices/Brain.js
+++ b/devices/Brain.js
@@ -1,0 +1,108 @@
+/* Copyright (c) 2015 Dennis Bemmann. See the file LICENSE for copying permission. */
+/*
+Brain.js
+========
+
+This module interfaces the NeuroSky ThinkGear ASIC module (TGAM) for use with Espruino.
+
+The module returns the following information:
+- signal quality
+- attention value
+- meditation value
+- EEG power band values for the following brain waves as 24 bit unsigned integers:
+  delta, theta, low-alpha, high-alpha, low-beta, high-beta, low-gamma, mid-gamma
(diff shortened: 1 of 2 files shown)


gfwilliams:
Great - love it! In case anyone else sees this, forum post is here too: http://forum.espruino.com/conversations/266808

pastaclub:
I noticed that the module doesn't appear here:
http://www.espruino.com/Modules
where it probably should be listed.

I also noticed that the layout of this page
http://www.espruino.com/Brain
is broken. I committed a markdown file which looks fine but that html page is scrambled somehow.

gfwilliams:
Thanks - I don't know why but the Markdown parser I use (marked) doesn't seem to like things when they don't have a newline between them. I actually [fixed it](https://github.com/espruino/EspruinoDocs/commit/8dadf6fd6ef8847f5bfe36cbe430eec3a80848c8) a few days ago but forgot to update the site with the new version. I've done it now though.

PR #136, opened 2015-04-29
Corrected mistakes in example

pastaclub:
1) f is undefined, this should have been logFile.
2) If the button is supposed to toggle the operation (as stated above), the function must only be called on one edge, not both.


gfwilliams:
Thanks! My mistake for dragging off the forum without testing!

PR #140, opened 2015-05-13
CAP1188 breakout module

andrewn:
This is a port of the Adafruit CAP1188 breakout board Arduino library.

I've kept it simple, this module just allows the 8 touch channels to be read. However, the CAP1188 itself can do a lot more so there's scope to add more functionality later.

I've tested this on an Espruino v1.3b board and the Pico (both with 1v78 firmware).


gfwilliams:
That looks great - thanks! I'll stick it on the site when I next do an update.

andrewn:
Great, thanks!

PR #143, opened 2015-05-17
fix typo in example

SpenceKonde:


gfwilliams:
:o oops. that was quite a nasty one. Thanks!

PR #146, opened 2015-06-04
Added i2c address as connection parameter

gomako:
Hey, this is my first pull request so apologies if I'm doing it wrong! I saw [this thread](http://forum.espruino.com/conversations/263400/#comment12127289) about specifying an i2c address for different 1306 modules and needed to do it myself. Let me know if the changes are ok and it's all ok?

### DIFF
--- a/devices/SSD1306.js
+++ b/devices/SSD1306.js
@@ -12,7 +12,7 @@ function go(){
 
 // I2C
 I2C1.setup({scl:B6,sda:B7});
-var g = require("SSD1306").connect(I2C1, go);
+var g = require("SSD1306").connect(I2C1, 0x3C, go);
 
 // SPI
 var s = new SPI();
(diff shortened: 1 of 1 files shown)


gfwilliams:
Hi Ben,

Thanks! It looks good - but unfortunately the changes will cause problems for people who're already using it, so ideally we'd have something that added the functionality (and maybe the ability to take other info too) while still being backwards compatible. Could you try something like:

```
exports.connect = function(i2c, callback) {
  var oled_address = 0x3C;
  if ("object"==typeof callback) {
   if (callback.address) oled_address = callback.address;
   callback = callback.callback;
  }
```

you could then:

```
require("SSD1306").connect(I2C1, go);
// or
require("SSD1306").connect(I2C1, {address:0x3C, callback:go});
```

If you update your own repository with those changes, I think this pull request will automatically update, and I can merge it in.

gomako:
Ah of course! My javascript is fairly weak. I will update the code and push it!

gfwilliams:
Sorry for the delay getting this in!

PR #147, opened 2015-06-11
Update broken link to image

peterhgombos:
The link to the image was broken, now it is not.


gfwilliams:
Hi Peter - Does the link to the image not work for you on the actual Espruino site? http://www.espruino.com/Power+Consumption

That's were it's expected that you'll view these pages.

It's actually an issue with all the markdown pages in this repo, not just this one - the tool I use to turn them to HTML actually assumes the path includes a directory of the same name as the page, which works fine but it means that the images themselves aren't visible when viewed on GitHub.

peterhgombos:
Ah, it worked there, I just got here from one of those sites. And it was a lot easier to browse the different pages over here. If there's no way of making it work both places, disregard this.

gfwilliams:
So you got here from Espruino.com? Out of interest, why was it easier to browse on GitHub? If the Espruino site is hard to use somehow I'd like to fix it :)

I'll leave this open and will see how much effort it is to swap everything over to do images this way. It would be nice to have *both* GitHub and espruino.com displaying everything properly.

peterhgombos:
I can't find an index of these files, and there's no easy way to find the way into one of these pages.  When I click on Documentation from the front page, I can't find an index of the docs, neither in the drop down or in the left-hand menu.

gfwilliams:
Closing this now - I think I've updated *all* images to the correct link style

PR #149, opened 2015-07-21
Add reset method to CAP1188 module

andrewn:
This enables use of the reset pin on the CAP1188 breakout.


gfwilliams:
This looks great - thanks!

PR #151, opened 2015-07-30
Update Pyroelectric.md

SpenceKonde:
I just spent an hour trying to figure out why this wouldn't work.

Turns out the delay was just cranked to the max, and the pulse was so long I assumed it was active low and not triggering (yeah, the datasheet says otherwise, but considering the engrish, I put little stock in it). I was dancing around in front of it, trying different supply voltages, and nearly posted to forums asking for help.


gfwilliams:
Great - thanks! Frustrating about your unit - I thought they were supposed to be 'calibrated' to ~2 sec in the factory.

PR #153, opened 2015-08-11
TCS3200 - initial commit

darkmoon32:
Initial commit for TCS3200 modules. Code was tested on YL-64 module and is able to detect red, green and blue color


gfwilliams:
Great, thanks! I don't have one here to test with, but do you think I could make a few changes and then you could test and make sure it still works?

I think it'd be really handy to have `getValue` as well, a bit like on here: http://www.espruino.com/TCS3472x

People might want to figure out hue in more subtle ways.

gfwilliams:
Ok, just tweaked - please can you try:

```
var sensor = require("https://raw.githubusercontent.com/espruino/EspruinoDocs/master/devices/TCS3200.js").connect(A0, A1, A2, A3, A4);
setInterval(function(){
  var c = sensor.getColor();
  digitalWrite(LED1, false);
  digitalWrite(LED2, false);
  digitalWrite(LED3, false);
  if(c == 'red')
    digitalWrite(LED1, true);
  else if(c == 'green')
    digitalWrite(LED2, true);
  else if(c == 'blue')
    digitalWrite(LED3, true);
  //print(c);
}, 200);
```

and see if it still works? When we're sure it does I'll stick it on the Espruino website and you'll be able to access it with a normal `require('TCS3200')`

darkmoon32:
Well it still work, but I had to make one fix on line 46 from column 16 to column 20. There shouldn't be "tcs." so green was undefined

gfwilliams:
Great, thanks! I'll push it up to the site this week

PR #154, opened 2015-08-18
RGB Led module

bartmichu:
Just a small helper module for beginners.


gfwilliams:
That's great - thanks! Sorry for the delay in merging :)

PR #156, opened 2015-08-31
Bosch Sensortec's BME280 temperature/pressure/humidity sensor

mokjpn:
I have wrote a module to read data from Bosch Sensortec's BME280 temperature/pressure/humidity sensor.
This can be used for the simplest usage to obtain the sensor's value. There is no option to change modes of sensor yet.


gfwilliams:
That's awesome - thanks! I'm on holiday at the moment, but will put this on the website next week.

PR #159, opened 2015-09-13
Modification on BME280 module's instruction.

mokjpn:
I slightly modified instruction of BME280 module.

- When you use I2C, you have to connect CSB pin to 3.3V.
- The module uses default I2C address (0x76), so you have to connect
SDO pin to GND.


gfwilliams:
Thanks!

PR #161, opened 2015-09-14
fix gain setting in conversion

allObjects:
To set gain in register on convert, use ‘value’ (from ‘table’ object GAINS) and not just the ‘pointer’ (this.gain).


gfwilliams:
That's great - thanks! Embarrassing mistake ;)

PR #162, opened 2015-09-16
Add module support for MCP23xxx port expanders

SpenceKonde:


gfwilliams:
Wow, thanks! I'll stick those on the site this week - I'll quickly add a mention of the exact module names though, as I'm not sure the .md file says what you need to `require` for the 4 different types.

gfwilliams:
Ok, just done - there were some issues minifying so I [changed some lines](https://github.com/espruino/EspruinoDocs/commit/f94d36ae91aa8d03ad223226aa387c9c62f21b14) hope that's right?

Looking at the MCP23008 datasheet it's got an IRQ line. It'd be possible to actually do: `pin.setWatch(function)`. You could then internally setWatch the IRQ line, and when called you could then figure out which pin had changed state and call the appropriate function. I bit much, but it'd be pretty cool :)

SpenceKonde:
Thanks for cleaning that up - I realize the .md wasn't as polished as it could be - I'd had this code kicking around, not submitted or tested for a few weeks, and I just wanted to get it out there so people can use it. 

I was looking at the interrupts too - I may add that functionality at some point (or someone else may), but I wanted to get a working module into people's hands for these. 

I don't like pin.setWatch() particularly - setWatch() is the name of a builtin function. IMO better to have a different but related name, like pin.watch() (just like it's pin.read(), not pin.digitalRead() )... not sure what the wider perspective is on this though.

gfwilliams:
No problem! Thanks for submitting the code anyway - I'd much prefer to have the code available with less than perfect documentation. As long as there's a snippet of example code there are plenty of people that won't read the text anyway and will dive right in :)

`pin.watch` makes a lot of sense - but yeah, that's pretty low priority really. Most people will just want to be able to use the IO expander.

PR #164, opened 2015-10-11
CH_PD requires 3.3v

olliephillips:
Seemed as quick to send a PR as explain. Table that confused amended in this PR.


gfwilliams:
Awesome - thanks!

Yeah, sorry - I can see how that's confusing :)

PR #165, opened 2015-10-12
Update to account for new devices

SpenceKonde:
Some CBRAM is spec'ed with much higher endurance. 

There's a 1.5v logic level through-hole mosfet I missed last time. 50 Vds too!


gfwilliams:
Great - thanks!

PR #172, opened 2015-10-15
minor correction

williammcrae:
I am seeing the following error using the MPL115A2 module:
  WARNING: Expecting a number or something iterable, got undefined
Needed a minor correction.


gfwilliams:
Thanks! Wow, not sure how that slipped through!

PR #174, opened 2015-10-17
Added docs for esp8266 board

tve:
I'm not sure how the photo for the esp-12 module gets pulled in from the Espruino repo and I also don't know where to insert the esp-12 into the list of boards being displayed.


gfwilliams:
looks perfect - thanks! will try and update the site with it next week

PR #175, opened 2015-10-17
Updating API and adding WebSocket options

samehhady:
We are now using the same API as the node.js npm module.
The WebSocket page will be updated to reflect the new changes.


samehhady:
I am not sure about the first commit, What I did is that I synced my fork with the latest updates then done my updates

PR #177, opened 2015-10-19
Fixing ws issue that didn't appear till it was delivered

samehhady:
I can't call new Websocket directly without including a function, while it was working fine while testing using var exports = {};
need to discuss it more with Gordon


gfwilliams:
Hi Neil, if you'd just closed this PR in the first place I think it would have been a lot more clean than merging it, and then merging something that reverted it.

PR #180, opened 2015-10-21
add info on console over serial to serial port discussion

rmcubed:
Made a bit of a mess of the commit history, but there's only one update in there. I see @gfwilliams committed the same info to Troubleshooting, so maybe this change is redundant.


gfwilliams:
Thanks - the more places it's mentioned the better really. I'm surprised I didn't have something about it before.

PR #182, opened 2015-10-25
Espruino module for interfacing MCP9808 temperature sensor

luwar:
This is a pull request to add support for the MCP9808 temperature sensor from Microchip. The implementation is feature-complete and tested. The documentation explains only the basic usage (getTemperature()) but lacks advanced use cases such as alerting by temperature limits or wakeup/shutdown mode.
My English knowledge is very limited. It would be nice if a native could review the sentences in the JavaScript and the Markdown file.


gfwilliams:
This looks awesome - thanks! Sorry for the delay, I've been away so far this week.

PR #185, opened 2015-10-28
added emw3165 board

tve:


gfwilliams:
Thanks - although right now there's not even an EMW3165 board file in the repo?

PR #187, opened 2015-11-02
Encoder module: Pins will be checks during initialization and module state will be calculated correctly.

luwar:
The encoder module did not check the states of the two input pins. this.last was set to 0 while PINA and PINB could be 1.
This leads to the strange behavior that the first rotary movement could result in +1/-1 while the following movement will result in +2/-2.

This pull request uses the available onChange() method to initialize the state of the module without changing any other behavior.


gfwilliams:
Thanks!

PR #190, opened 2015-11-18
Created package.json file & Updated Read Me

chalkers:
Added a package.json file to help handle JavaScript dependancies better.

Updated readme for the Espruino source requirement.

Included instructions for using npm as a task manager. i.e. `npm run build` & `npm start`.


README.md
```diff
@@ -1,10 +1,11 @@
 <!--- Copyright (c) 2013 Gordon Williams, Pur3 Ltd. See the file LICENSE for copying permission. -->
-EspruinoDocs
-============
+# EspruinoDocs
```
gfwilliams: Why did you change all these? Some of us like underlining the titles ;)

README.md
```diff
@@ -16,76 +17,100 @@ This is basically [GitHub Markdown](https://help.github.com/articles/github-flav
...
+## Build Process
 
-Finally you need to install a node module for highlighting:
+### First Time:
+
+For the first time you can run:
 
 ```
-$ npm install marked highlight.js acorn express tern --save
+$ ./build.sh
```
gfwilliams: This might not be entirely right. For instance if you want your modules to be updated, or the pinout for each board to be updated, you'll still need build.sh I think?

gfwilliams:
Thanks! Just merging this in. Made some comments though ^^^

README.md
chalkers: I think they still are :)

README.md
chalkers: Is that what this code does in the bash script?
```
python scripts/build_board_docs.py PICO_R1_3 pinout
mv boards/PICO_R1_3.html $DIR/html/boards
python scripts/build_board_docs.py ESPRUINOBOARD pinout
mv boards/ESPRUINOBOARD.html $DIR/html/boards
python scripts/build_board_docs.py ESP8266_BOARD pinout
mv boards/ESP8266_BOARD.html $DIR/html/boards
python scripts/build_board_docs.py EMW3165 pinout
mv boards/EMW3165.html $DIR/html/boards
```

What are the scenarios where you need to do this?

Also not too keen that that shell script randomly generates this folder - `WEBSITE=~/workspace/espruinowebsite` lol

Oh and there's this:
```
bash buildmodules.sh
```
Is that where you mean build modules.

So what's the main differences between bin/build.js and build.sh?

gfwilliams: Yeah, so the python generates something that looks like this: http://www.espruino.com/ReferenceSTM32VLDISCOVERY

And then that can get pulled into a markdown file like this: http://www.espruino.com/Pico

And yeah, `~/workspace/espruinowebsite` is where I have the source for the actual Espruino site. Simplest thing would be to just check for its existence and then refuse to copy modules if it's not there.

And yes `bash buildmodules.sh` checks for updated versions of each module and minifies any that have changed.

README.md
gfwilliams: Yeah, they are the same - except *every other heading in the entire repo uses dashes*, so it feels a bit pointless to randomly change these to something different?

README.md
chalkers: Just to clarify:

bin/build.js is purely for the documentation (minus the boards with pinouts)

build.sh has 3 tasks:
1. generate pinouts using python
2. build the production site
3. minifies the modules

If that's correct I'll either update the docs tomorrow or create some other tasks in JS and use npm to run those tasks :)

gfwilliams: Yes, sounds about right. Thanks!

I reckon `build.js` could call Python to generate the board docs directly (it'd be more sensible) - it was just me being lazy using bash for it.

The module minification is an important one, because it's very handy to be able call it separately if you're making changes and want to quickly see how the minified code comes out.

PR #193, opened 2015-11-24
Conform to NodeJS MQTT.js API

heri16:
Amazon Web Services IoT uses the mqtt nodes library from https://github.com/mqttjs/MQTT.js . Changes are made to conform to this API (while maintaining backwards compatibility for now). This makes nodejs libraries easier to port to Espruino.


gfwilliams:
Great - thanks! Is it worth printing 'deprecated' for the old function calls we don't want people using?

Also, the docs (`MQTT.md`) should really be updated to match the new API.

PR #194, opened 2015-11-24
Port of aws-iot-device-sdk /device/index.js

heri16:
This port works with new MQTT.js API here: https://github.com/espruino/EspruinoDocs/pull/193


gfwilliams:
Thanks - this could really do with some documentation though. Could you add a `.md` file? For instance presumably 'region' has to be set to something in the options, but I have no idea what.

PR #200, opened 2015-12-10
Add ST-Link Utility step; erase before program

AlexanderBrevig:
A few times I've had the weird issue that the Espruino stops connecting to USB. It ceased when I started erasing before programming. 

Is there somewhere else I should add this tidbit? I did try and search google with some terms using `site:espruino.com` but this is the only mention of ST-Link Utility  that I could find.


gfwilliams:
Thanks!

I think that's the only one. There's http://www.espruino.com/AdvancedDebug but generally I suggest that people just use `make flash` which contains the right command (and offset) - it's a lot easier :)

PR #203, opened 2015-12-13
Support of all HTU21D sensor features

luwar:
I have completed the HTU21D module to support all sensor features:

- Get and change sensor resolution
- Read temperature in blocking/synchronous mode
- Read humidity in blocking/synchronous mode
- CRC Checks added
- Read chip identification / serial number
- Calculate temperature compensated humidity
- Calculate the dew point
- No "console.log" for standalone operation
- Heater support
- Soft Reset
- Read End of Battery Status

There is no api break. All functions were testet with a real sensor.


gfwilliams:
Looks great, thanks!

PR #209, opened 2016-01-03
Module for MMC212xMG (Dual-axis Magnetic Sensor)

luwar:
Hi Gordon,
I made a module for MMC212xMG, a dual-axis magnetic sensor from MEMSIC with I²C interface. The sensor is not very common anymore but still available.
Could I merge the pull request? Thanks :-)
Luwar


gfwilliams:
That's great - thanks!

gfwilliams:
Hi - I just committed some changes that pull the initialisation of the object out of the function, and make it more like existing modules. Hope that's ok?

It doesn't seem to hurt minification much, but it means that when the `.md` file does `APPEND_JSDOC`, the script is able to parse out all the documentation you've written at the head of each function.

I'll try and update the website with it later today.

luwar:
Thanks for accepting the module. It's all right that you changed the initialisation of the object. I'm programming in TypeScript and this IIFE pattern will be applied automatically to avoid global namespace pollution. I will avoid it in the next modules.

But there is still a encoding problem on the website: http://www.espruino.com/MMC212xMG
For example: "at 25Â°C" instead of "at 25 °C".

This is the same behaviour in other modules, e.g. http://www.espruino.com/MCP9808

It's causes by
```
<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1" />
```

With
```
<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
```
all things look fine.

Could you change the encoding in the template of the website?

gfwilliams:
That's great, thanks! I was meaning to fix it, but hadn't got around to investigating what was wrong... I've just changed it.

PR #210, opened 2016-01-06
Fix DHT22 lib for AM2301

AkosLukacs:
@gfwilliams I only have an ESP8266 board, so can't verify it on a real espruino and to test the issue you would need a AM2301. I hope i can explain this properly...
The problem: AM2301 is supposed to be compatible with DHT22, but (at least the one I got) transmits some "0" bytes after the real payload. 
An annoted screenshot from logic analyzer showing DHT22 and AM2301: https://files.gitter.im/espruino/Espruino/9vqG/blob
@SpenceKonde 's code didn't terminate after all the expected data received, and since additional bits received from the AM2301, that caused the checksum bit to be shifted (the exact number of additional shifts depends on timing, the code waits for 50 ms).
So modified the code to terminate after 40 bits received (of course, left the 50ms timeout in it).

Checked on EPS8266 with DHT22 and AM2301 side by side. The code still correctly handles if no sensor is plugged in.

@SpenceKonde please take a look at this, since you wrote the original. I have non-image exports from the logic analyzer those captures...

### DIFF
--- a/devices/DHT22.js
+++ b/devices/DHT22.js
@@ -6,6 +6,8 @@ Usage (any GPIO pin can be used):
 var dht = require("DHT22").connect(C11);
 dht.read(function (a) {console.log("Temp is "+a.temp.toString()+" and RH is "+a.rh.toString());});
 
+the return value if no data received: {"temp": -1, "rh": -1, "checksumError": false}
+the return value, if some data is received, but the checksum is invalid: {"temp": -1, "rh": -1, "checksumError": true}
   */
 
 exports.connect = function(pin) {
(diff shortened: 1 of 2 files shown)


SpenceKonde:
That looks solid. OFC, I don't have any of the offending units to check.

gfwilliams:
Looks great, thanks! Seems like it would work fine...

@SpenceKonde do you have a normal DHT22? If you got a chance to check this with DHT22 + STM32-Espruino it'd be really handy, but I'll merge this anyway for now.

PR #211, opened 2016-01-08
Update KeyPad.md

stephaneAG:
just passing by as I received those from ebay & have to build an easier TV remote for my gran' pa ;p (..)
"Tiny mistakes quickly fixed"


gfwilliams:
Thanks! Not sure how that got in there!

PR #213, opened 2016-01-18
Improve esp8266 docs

tve:
This also adds a wifi usage tutorial.


gfwilliams:
Thanks! By the way, for things like the tutorials, it might be worth adding some keywords? It'll create some based on the title so it might be ok.

Potentially you can also do what I do with the Pico - have `USES` and `APPEND_USES` to automatically add  list of tutorials that use a certain thing. So maybe `EspruinoESP8266` to avoid confusion with the existing stuff that uses the module via AT commands

PR #216, opened 2016-01-21
tried to update the 'connect' method, allowing to configure the brigh…

yerpj:
…tness of the display, through an option {brightness:value}, with value=[0..255]. Seems to work, but may not be linear

### DIFF
--- a/devices/SSD1306.js
+++ b/devices/SSD1306.js
@@ -135,7 +135,3 @@ exports.connectSPI = function(spi, dc,  rst, callback, options) {
   // return graphics
   return oled;
 };
-
-exports.brightness = function(x) {
-	
-};
--- a/devices/SSD1306.js
+++ b/devices/SSD1306.js
@@ -72,6 +72,18 @@ exports.connect = function(i2c, callback, options) {
 
   // configure the OLED
   initCmds.forEach(function(d) {i2c.writeTo(addr, [0,d]);});;
+  
+  //set the brightness once
+  if(options && options.brightness){
+    var x=options.brightness;
+    if (x>255)
+      x=255;
+    if(x<0)
+      x=0;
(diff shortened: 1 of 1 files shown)


gfwilliams:
closing this, as added a `setContrast` function

yerpj:
Thank you so much @gfwilliams ! I'll use this right now!

yerpj:
I just tested the new function ```setContrast()```. It seems that the command is misunderstood by the display. I wonder if we rather have to write this in two steps. I tried earlier with that one: 

```i2c.writeTo(addr, [0,0x81]);
i2c.writeTo(addr, [0,x]);
``` and it worked. 


Or maybe could we try ``` i2c.writeTo(addr, 0,0x81,0, c);```, I don't know...

gfwilliams:
Ahh, Damn - i only tried the SPI one.  But the `{contrast: 0}` option works?

What about `i2c.writeTo(addr, 0,0x81,c);`? You could try that by hand, specifying I2C1/whatever and see if it works?

yerpj:
```  g = require("SSD1306").connect(I2C2, function(){},{contrast:255}); ``` --> this works well

```  var value=3; //contrast value
I2C2.writeTo(0x3c,0,0x81,value);  ```  --> this works as well

gfwilliams:
Great - I'll update it now

yerpj:
@gfwilliams thanks a lot! Last question: Why are the files at http://www.espruino.com/modules not updated automatically based on the current git files? are they manually updated?

gfwilliams:
Yes, it's manual. Sometimes modules get updated in there but they rely on a new version of the Espruino firmware - so then I hold off and release both together

yerpj:
That make sense.  
Thank you for the update.

PR #218, opened 2016-01-27
Module for Infrared Thermometer MLX90614

luwar:
Hi Gordon,
I made a module for MLX90614, a infrared temperature sensor. It's a cool thing with a lot of possibilities for experimenting.
Could I merge the pull request? Thanks :-)
Luwar


gfwilliams:
That looks great... thanks! I'll stick it on the site when I next do an update.

PR #219, opened 2016-01-28
Change B0 to B3 for NRF24L01P

ForbesLindesay:
The Pico doesn't seem to have a B0 pin.  Changing to B3 seems to work fine so I think it should probably be the recommended option.

### DIFF
--- a/devices/NRF24L01P.md
+++ b/devices/NRF24L01P.md
@@ -19,7 +19,7 @@ Connect as follows:
 | 1 | GND       | GND        | GND |
 | 2 | 3.3V      | 3.3        | 3.3v |
 | 3 | CE        | B1         | GPIO Out |
-| 4 | CSN       | B0         | GPIO Out |
+| 4 | CSN       | B3         | GPIO Out |
 | 5 | SCK       | A5         | SPI SCK |
 | 6 | MOSI      | A7         | SPI MOSI |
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - unfortunately the issue is that all the wiring instructions (and pictures) are for the original Espruino board, where B3 is right on the other side of it. Changing that would mess this up for all owners of the original board.

Rather than changing that, it would be better to just have some kind of note (maybe link to the 'Notes' at the bottom from the table).

Of course ideally there would be a description of wiring specifically for the Pico... Since there's software SPI you can connect to whatever pins you want though.

There should probably also be a note that the [Shim pack](http://www.espruino.com/Shims) contains an adaptor to stick the NRF24 right on the end of the Pico...

ForbesLindesay:
I don't have the shim pack, and didn't feel like I needed it for wiring up the NRF24.  I'm happy to leave it to you to decide what to do here, just wanted to suggest that it could be improved.

gfwilliams:
Ok, thanks. I'll leave it open, and will make some notes about the Pico at some point.

ForbesLindesay:
:+1:

gfwilliams:
Now changed title in table, added note about having custom pins, and added links to the Pico NRF24 adaptor shim and pinout.

PR #221, opened 2016-01-30
New module NRF905 for 433MHz & 868/915MHz 

yerpj:
I recently tried to establish a wireless link between some of my devices, and I used those [sadly famous transmitters/receivers](http://www.robertoinzerillo.com/wordpress/wp-content/uploads/2012/10/RF-433MHz.jpg).

It appears that the receiver is totally useless, due to its simple super-heterodyne structure and the lack of output filtering. 
Then I had a look at some not so cheap, but cheap enough [rf modules based on the NRF905](http://www.dx.com/p/nrf905-wireless-communication-transmission-module-for-arduino-148058#.VqwWg1JN8Vc) .

Basically, the NRF905 is more or less similar to the NRF24L01, unless it can transmit on sub-GHz bands.
I tried to implement a basic driver for it, and I hope, if the community is interested in having such a module, that people will have a look at the code, because I am not very comfortable with the Object Oriented style. I am sure there is a lot of code to clean.

ps: I am sorry, I was not able to generate a pull request ONLY on the NRF905.js file. Please, do not take into account the other file SSD1306.js.

### DIFF
--- a/devices/NRF905.js
+++ b/devices/NRF905.js
@@ -177,7 +177,7 @@ NRF.prototype.init = function(myAddr,Callback) {
 	this.setReg(REG.W_CONFIG+2,0x11);	//both TX and RX addresses are 1 byte long
 	this.setReg(REG.W_CONFIG+3,0x20);	//RX payload is 32 byte long
 	this.setReg(REG.W_CONFIG+4,0x20);	//TX payload is 32 byte long
-	this.setReg(REG.W_CONFIG+9,0x58);	//8-bit CRC enabled, Fosc=16MHz, disable external clock
+	this.setReg(REG.W_CONFIG+9,0xD8);	//16-bit CRC enabled, Fosc=16MHz, disable external clock
 	digitalWrite(this.TRXEN,1);//enable radio
 	digitalWrite(this.TXE,0);//enable RX
 	var a=this;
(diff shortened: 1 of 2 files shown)


gfwilliams:
Hi - sorry for the delay...

Thanks! I just pulled this in manually. I'll be on the website next time I update.

I tweaked the formatting a bit, as generally the modules use K&R style. The code looks fine though - generally I'd set unused callbacks and things to `undefined` and then just check them with `if (mycallback)` but it's not a big deal.

I'm afraid the NRF24 module is a pretty nasty one to copy off. It really needs a lot of polishing :)

Also, I added a simple NRF905.md for documentation - it'd be cool if maybe it could have a bit more info in, but I don't have time for that right now.

yerpj:
Thank you very much! I will work on that in the future. As far as I know, there is still a few bugs in the way the NRF905 is driven. I was able to receive about 70% of the packets transmitted, I would like to rise this ratio up to 95% at least. I will also work on a link layer, in order to provide some kind of automatic acknowledgment. I will also describe the driver behaviour in the NRF905.md.

I will be in holiday until end of february, therefore all that stuff will begin in early march.

gfwilliams:
Just to add - it'd be really cool to (eventually) make all of these packet radios (nRF24, nRF905, RFM12, RFM69, ...) have a similar kind of API for starting reception and transmitting packets.

There's a lot of code that could be useful to all of them, like listening for only some of the time, or simple meshing, that could then be shared.

PR #223, opened 2016-02-07
Espruino Module for Magnetometer MAG3110

luwar:
Hi Gordan, I made a new module for MAG3110, a magnetometer from NXP/Freescale which is also built into the Micro:bit. A few comments:

* I put a self-made photo into the documentation. It's not good but I wouldn't use any from Ebay. Is it legal to take a nice picture from Ebay? What do you think?
* It would be nice if microbit user can quickly start with the MAG3110 module. Could you add a line or two in the documentation (Usecases → Initialisation) which pins should be used in the connect()-method? I would do it by myself but can't buy a microbit. Where do you get one?
* The module is JSDoc-documented with Closure Compiler annotations and ADVANCED_OPTIMIZATIONS in mind. But I didn't enable @compilation_level ADVANCED_OPTIMIZATIONS. I dislike the idea of using array notation (mag['getMode'])() instead of mag.getMode(). The BME280 module use this, e.g. this.i2c['writeTo'](). It makes the code less readable and IMO this is a step in the wrong direction. Better would be to tell the Closure compiler not to minify externs and export code:
https://developers.google.com/closure/compiler/docs/api-tutorial3#no 
Probably this is not an easy task: The complete Espruino api (which is available for modules) have to be written in a file extern.js (without implementation) and must be used while minifying. I don't know how difficult it is to adapt the build process.

Hoping you accept the PR. Tell me if something is missing.


gfwilliams:
Nice, that's great - thanks!

I probably wouldn't use an eBay image - I don't want to get anyone complaining about copyright... I've tried only to use other people's images when I actually know them and can ask. Could you take a photo of it on a white sheet of paper? It's then pretty easy for me to fiddle with the contrast to give it a white background?

For the micro:bit there's actually a [simple bit of code for it built-in](https://github.com/espruino/Espruino/blob/master/libs/microbit/jswrap_microbit.c#L217) which I'd probably encourage micro:bit users to use (there's so little RAM available (~300 vars) I didn't want to fill it up with modules). However it doesn't handle any kind of calibration at the moment :(

Maybe as you've looked into it a bit more you'd have some ideas about what to do to properly set it up for people to use (just taking very basic readings).

At the moment micro:bits still aren't available, it took me months of trying to contact people before I got one. Hopefully they'll be available for sale in a few months though.

PR #225, opened 2016-02-29
Update SSD1306.md

MaBecker:
I2C: add rest pin

### DIFF
--- a/devices/SSD1306.md
+++ b/devices/SSD1306.md
@@ -112,6 +112,17 @@ require("SSD1306").connect(I2C1, start, {
   contrast : 31,
 });
 ```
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - will put this in later.

However I'll need to tweak it a bit - when displays use a reset pin it's usually called `rst` and `reset`, not `pin`.

Also, this will only work on ESP8266. Because the ESP8266 isn't good at accurate timings, `digitalPulse` blocks on that, but not on anything else. The result would be that on any other device, reset actually resets the board while it is uploading config info.

MaBecker:
cool

PR #227, opened 2016-03-09
Added address parameter to specify PCF8574 backed LCD location

iRet:
It would be good to receive custom pins as well but I have no idea how deal with all this bitmask mess.
So at least address for now. It solves my problem. May it would be useful for someone else.


gfwilliams:
Nice, thanks! that's really tidy!

For more complex things, I guess it's worth just adding `exports.HD44780 = HD44780;`, then someone can just make their own with their own `write` function?

gfwilliams:
I'll pull this in but it won't go on the main website for a few days...

PR #228, opened 2016-03-15
Better error handling with clearer messages

chalkers:
Added return codes + messages for better error handling. Less of a black box now :)


gfwilliams:
Great - thanks!

PR #229, opened 2016-03-25
ADVANCED_OPTIMIZATIONS support with externs files added

luwar:
I implemented advanced optimization with externs files, mentioned in http://forum.espruino.com/conversations/282737/

* The new feature is shown in module BME280: Although already advanced optimized with array syntax the new code is even 7% smaller.
* The minified code of all other modules have not changed.
* Compiler warning will now be shown in the console and not suppressed anymore.

My next steps will be:

* Writing externs files for all modules I can test.
* Starting unit testing of modules: First I will write my ideas to the forum.
* Generating espruino.externs from espruino.json

### DIFF
--- a/bin/espruino.externs
+++ b/bin/espruino.externs
@@ -1,0 +1,93 @@
+/**
+ * @constructor
+ */
+function Graphics() {}
+Graphics.createArrayBuffer = function( width, height, bpp, options ) {}
+Graphics.createCallback = function( width, height, bpp, callback ) {}
+Graphics.createSDL = function( width, height) {}
+Graphics.prototype.getWidth = function() {}
+Graphics.prototype.getHeight = function() {}
+Graphics.prototype.clear = function() {}
+Graphics.prototype.fillRect = function( x1, y1, x2, y2) {};
+Graphics.prototype.drawRect = function( x1, y1, x2, y2) {};
+Graphics.prototype.getPixel = function( x, y ) {};
+Graphics.prototype.setPixel = function( x, y, col) {};
+Graphics.prototype.setColor = function( r, g, b) {};
+Graphics.prototype.setBgColor = function( r, g, b) {}
+Graphics.prototype.getColor = function() {};
+Graphics.prototype.getBgColor = function() {};
+Graphics.prototype.setFontBitmap = function() {};
+Graphics.prototype.setFontVector = function(size) {};
+Graphics.prototype.setFontCustom = function( bitmap, firstChar, width, height) {}
+Graphics.prototype.drawString = function(str, x, y) {};
+Graphics.prototype.stringWidth = function( str ) {}
+Graphics.prototype.drawLine = function( x1 , y1, x2, y2) {};
+Graphics.prototype.lineTo = function(x, y) {};
+Graphics.prototype.moveTo = function( x, y) {};
+Graphics.prototype.fillPoly = function( poly ) {};
+Graphics.prototype.setRotation = function( rotation, reflect) {};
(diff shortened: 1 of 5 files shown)


gfwilliams:
Thanks - as I'd said in the forum post, I'm still a bit worried about this - especially as the `externs` list is incomplete (and if manually created is guaranteed to become out of date very quickly).

The BME280 module itself looks a lot tidier now though. Is there any chance of getting something to build the externs file automatically? Something like this looks like it does almost the right thing already: https://github.com/espruino/Espruino/blob/master/scripts/build_jsstub.py

luwar:
Ok, it is a pity. I don't understand why it is worse than before.

* Without advanced optimization → nothing has changed.
* So we talk only about advance optimization:
Currently everybody have to take care to replace all i2c.writeTo/readFrom or module.property calls with i2c['writeTo/readFrom'] and so on. This is really error-prone and there is no help from the editor/IDE.

I know that's not perfect but it would be better than before.

If anybody could find a method to extract an externs file from the module code that would be really cool. But I don't believe that it is possible. There are so many similar questions on Stackoverflow with no real solution. One approach is to execute the library code and then extract the properties/methods of an entry object, e.g.
   http://blog.dotnetwise.com/2009/11/closure-compiler-externs-extractor.html

The other approach is to export functions which are annotated with @export and replace the goo.exportXxx function calls. But therefore the Closure Compiler Application is required. The Web API doesn't provide this functionality.

I don't know any library which have autogenerated externs files. I don't have ideas anymore.

gfwilliams:
I'm not worried about extracting the extern file for each module... It's not that hard to do as you say.

The problem really is the main Espruino one. It's not complete, and nothing is documented about how you use it. When this PR goes in, the changes might be useful to you, but for everyone else I don't think it's that easy to see how to use it (most people don't even try and minify their modules before submitting them).

I'll merge this anyway - as you say the BME280 module is a lot nicer with it. I just don't expect that anyone will use this as-is without any documentation and with the Espruino exports only really being enough to get your one module compiling.

luwar:
Ahh, I misunderstand you, sorry. I thought you want to generate the module externs files and that's hard.

The espruino.externs has to be generated. I wrote this in step 3 of my first post. I will do this next. Is the file espruino.json the right place to start the converting process?

What is your preferred tool for doing this? Different parts of the build use different tools: JavaScript/node, Python, Bash or Make (for Espruino). In which language should I write the converter?
BTW: Why are so many tools involved?

Just for explanation: I made the PR to avoid one huge commit without any feedback from time to time. It's clear that the current state is not ready to be promoted but it's still a bit better than before, IMHO.

Thank you for merging. :-)

gfwilliams:
Personally, I'd use: https://github.com/espruino/Espruino/blob/master/scripts/build_jsstub.py

This does almost exactly what you want already, direct from the Espruino source code. Something very similar updates espruino.json and I can modify things to update `espruino.externs` at the same time.

... Of course if you're using `espruino.json` then you could auto-create `espruino.externs` in `minify.js`?

Personally I'd use node.js as it's needed for the minification anyway.

Honestly Make is needed, and Bash is used just for very basic scripting. I started off using Python for more serious scripting (like making the symbol tables) as it comes pre-installed on Linux but I find it extremely irritating, so for newer things, when I get the chance, I use node.js.

As much as I dislike writing Python, I think I'm stuck with it if we want to be able to compile on very constrained platforms like OpenWRT. About the only other solution would be to compile a 'bootloader' Espruino that didn't need any scripts (pre-built symbol table) and to then use that for building everything else.

PR #235, opened 2016-04-15
Fix the writeBits() method

trandi:
This doesn't seem to change the raw Gyro or Accelerometre data that one gets, but I'll need this fix for he DMP configuration that I'm planning on adding to this module.

Thanks,
Dan


gfwilliams:
Thanks! Looks good...

PR #236, opened 2016-04-15
Some MQTT fixes.

rovale:
Fixed the error handling.
Fixed subscribing using an object.
Removed some unnecessary semicolons.
Fixed some minor issues with variables.


gfwilliams:
Thanks!

PR #237, opened 2016-04-21
Add support for Digital Motion Processor, update docs

trandi:
Here it is, it would be really unfair for this module not to use its DMP feature, as it's non trivial to do the integration and filtering in code.
My only worry is that the new MPU6050_DMP is very big (~13kB minified) and it barely fits in the Pico's memory. All due to the big blobs of binary data (the actual code and config that it will run) that need to be written to the module at config time.

Any feedback is welcome !


trandi:
@ltj could you please review as the author of the original MPU6050 module ?
Thanks,
Dan

trandi:
@gfwilliams any chance this PR gets approved ? Any feedback on the changes ?

gfwilliams:
Wow, sorry for the delay - not sure how I missed this.

Looks great - could you make DMP_MEMORY and DMP_CONFIG into `Uint8Array` though? Like `var DMP_MEMORY = new Uint8Array([ .... ])` - that will drastically reduce the amount of memory used :)

Also, can you change the instances of `const` back to `var`? `const` isn't supported by Espruino yet, and only works because it's actually treated as a variable:

```
>const X
Uncaught ReferenceError: "X" is not defined
>var X
=undefined
>const = 5
=5
>const
=5
```

gfwilliams:
Just to add - `const` is an ES6 thing, not ES5 as far as I can tell.

gfwilliams:
One more thing - sorry.

Can you remove the following, unless there's a really good reason?

```
MPU6050.prototype.C = C;
MPU6050.prototype.R = R;
```

Basically, by making those public, you're stopping minification from inlining or getting rid of the unused fields of those objects. It makes a massive difference to code size.

trandi:
Hi Gordon, thanks for the quick reply !
However I'm not convinced by your argument around "const".
The example you give 
">const = 5
=5"
is not at all the same thing as the below : 

```
>const c = 5
=5
>c
=5
> 
```

Am I missing something obvious ??

trandi:
Also, when using new Uint8Array() for the 3 DMP_xxx lists, I don't see any improvement in the minified size, with or without the 'const' or 'var':

Original: 30422 -> 12724
With new Uint8Array() : 30464 -> 12766

What am I missing ?

trandi:
Same thing with 

```
MPU6050.prototype.C = C;
MPU6050.prototype.R = R;
```

I don't see any reduction in the minified size with or without them:
15316 -> 7550
15266 -> 7506

I must be doing something wrong I assume ? :)
The sizes I quote above are what appears in in the little blue pop up boxes when you press the "Send to Espruino" button...

gfwilliams:
You should probably be looking at `process.memory().usage` - while the source code is the same size, it's not the source code that is stored in Espruino - the source is executed and the result is stored. The result of `[1,2,3,4,5,6,7]` is much larger than `new Uint8Array([1,2,3,4,5,6,7])` - you can check with `E.getSizeOf([1,2,3,4,5,6,7])` (the difference is WAY more promounced for larger arrays - check http://www.espruino.com/Performance)

With `const = 5` my point is that Espruino treats const as a variable, not a reserved word. `const c = 5` is basically like writing `const;c=5` in Espruino (it seems it's kind of a bug that it doesn't error, but JS has some pretty weird semicolon rules). Basically `const` is an ES6 thing, so you probably shouldn't use it on the web either (not if you intend to target any older web browsers).

Maybe the lack of size difference is because you're using the Esprima minification? When actually putting the modules online I use the Closure Compiler, and that probably does some smarter optimisation.

I didn't realise you were using it from MPU6050_DMP - yeah, that makes a difference. It looks to me like someone who is using MPU6050_DMP won't be using many of the functions in MPU6050 though? I think in order to get RAM usage down to the point where it's actually usable, it's probably better to make the two modules entirely separate and duplicate the register and constant values.

It's a shame to duplicate stuff, but honestly if the module is so big you can't get any other code in, it's not going to be much use to anyone.

trandi:
Thank you Gordon for all the explanations.
I've integrated all your recommendations and pushed a new commit.
I can't make the 2 modules entirely separate as the _DMP one uses quite a few of the basic function in the original one.

Dan

P.S. there's one thing that still puzzles me in your response : "_it's not the source code that is stored in Espruino - the source is executed and the result is stored._ " is the opposite of what you say here http://www.espruino.com/Performance : "_while still keeping a copy of the source code it is executing so you can edit it on the device_" . What am I missing ?

gfwilliams:
Thanks! So when you upload something, Espruino does execute it - if you type `a=1+2`, `3` gets stored in a variable called `a`. However if you type `function a() { ... }` then a variable called `a` gets created that is a function containing the code `...` (as text).

So it is executed, but also the source is left around. It means that if you define an array, it's not the text of the array that is kept (unless it is in a function), but the variable containing the array.

gfwilliams:
Looks great - out of interest, what did the memory usage drop to?

Just wondering, for the second module you wrote - is it actually also based on Jeff Rowberg's code? If not, could I rip out that text and put the standard Espruino MPLv2 license on it?

trandi:
haven't measured actually, I just trust you that it's lower :)
Re the copyright, I've modified quite a lot the functions themselves but all the data in the array (which is actually the code that goes on the DMP's processor to do the filtering/integration) is exactly the same from Jeff's code, so I think we need to keep that bit.

But if you keep my name/copyright and the reference to Jeff Rowberg's code, the rest I guess you can replace ?

gfwilliams:
Ok, thanks - I imagine the code just comes from the chip manufacturers, but yeah, I think we'll have to keep that.

gfwilliams:
Just merged - I should get this on the website in the next update, probably early next week.

Thanks for all your work on this!

PR #242, opened 2016-05-12
wikipedia URLs are case-sensitive

The-Alchemist:
should be '#Key_frequency', not '#Key_Frequency'


gfwilliams:
Great - thanks! I'm pretty sure this worked in the past, so I kind of wonder if Wikipedia changed the case of the titles.

PR #249, opened 2016-05-26
Added instructions for Windows users

MrTimcakes:


gfwilliams:
That's great - thanks!

PR #256, opened 2016-06-08
formatting bug

The-Alchemist:
HTML renderer on espruino.com needs a space between ### and title


gfwilliams:
Thanks! Can't believe I missed that

PR #259, opened 2016-06-16
Improved documentation

mesutevin:
Re-placed a limitation warning in order not to be escaped the attention which led designing wrong pcb layout a few hours ago.


gfwilliams:
Thanks!

PR #261, opened 2016-06-23
fix STM32F1Flash.md example isPageSafe()

vjjft:
Though deprecated, the instructions are still useful for those working with the primary Flash module.  In 1v85 peek32() on 0xFFFFFFFF returns 2^32-1 which though bitwise equivalent to -1, is not equal to -1


gfwilliams:
Great - thanks for spotting that!

PR #263, opened 2016-07-23
Bugfix: HMC5883 module

mvaldesdeleon:
Hi, this is my first contribution to this project.

I was testing an HMC5883 3-Axis Digital Compass module and found some issues with the existing code.

With the suggested fixes, I am able to get readings as expected in both reading modes, and can confirm that the calculated values for the self-test are within the expected ranges.

Please let me know what you think, and if there is anything else to be done.


gfwilliams:
Sorry for the delay - I've been a bit busy recently.

This looks perfect though - thanks for all the fixes!

PR #264, opened 2016-08-11
initial check-in SSD1606.js module

programmicha:
please look over the SSD1606 module (doc and javascript)


gfwilliams:
Looks awesome - thanks for doing such a good document to go along with it as well!

PR #266, opened 2016-08-12
Issue 265 dcf77 return used watch

### LINKED ISSUE #265: DCF77 - return the used watch id
Right now the DCF77 module sets a watch with `repeat:true`. The user can't clear the watch other than with `clear();` which removes all watches.

It would be cool if the module return value (def object) contains the used watch id.

programmicha:
Fixes #265


gfwilliams:
Thanks! This did contain the SSD1606 stuff, but I think it was identical to your original PR so should be fine.

PR #269, opened 2016-08-23
Added Ping Module

MrTimcakes:


gfwilliams:
Great - thanks! I'll tweak the docs a little if it's ok, as this isn't really a proper 'ping' - even if it does basically the same thing.

PR #271, opened 2016-08-26
Add MySensors module

the1laz:
Still only does the basics and doesn't support connecting sensor networks yet, still only lets you interface sensors connected to your gateway. Should be useful for tying into IoT hub software though.


gfwilliams:
Great, thanks! I'm on holiday this next week so it won't go live for a bit, but I'll merge this into GitHub now anyway...

the1laz:
No problem, no rush :)

PR #278, opened 2016-09-09
Vvestm

VVESTM:
Hi,

Those 2 commits allow to get pressure and temperature from the sensors embedded in IKS01A1 ST expansion board.

Regards
Vincent

### DIFF
--- a/devices/HTS221.js
+++ b/devices/HTS221.js
@@ -84,7 +84,7 @@ function HTS221(_i2c, debugMode) {
 
 HTS221.prototype.ObtainCalibratedTemperatureCoefficient= function(callback) {
 
-  // T0_OUT, T1_OUT, T1_degC and T0_degC coefficients are factory calibrated
+  //  this.T0_OUT,  this.T1_OUT, this.T1_degC and  this.T0_degC coefficients are factory calibrated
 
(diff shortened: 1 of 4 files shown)


gfwilliams:
Looks great - thanks!

Only things I can see:

* The blank line on line 2 of the `.md` files causes my parser grief. It's no big deal for me to change, but if you're adding extra stuff it'd be handy if it didn't have that.
* You're using globals for T0_OUT/etc. That's fine for one device - but if you have two devices plugged in then those are *the same globals* in each case, so calibration may get messed up? To fix it you could use `this.T0_OUT` instead.

VVESTM:
OK Gordon, 

Thanks for this fast answer, I will push a new set of patches to take your remarks into account.

Vincent

gfwilliams:
Great, thanks!

VVESTM:
Thanks,

This will ease my demos and can be useful for other users getting the same hardware.
I guess it will be available through the web IDE in a few days ?

Regards
Vincent

gfwilliams:
Yes, the update is manual. As it happens I was updating some other stuff as well, so it's live now!

VVESTM:
effectively, I have checked and it is working !
good reactivity !! Have a good day !

PR #281, opened 2016-09-20
Updated Websocket module

AverageMarcus:
* Correctly handles connection upgrade without closing connection
* Allows sending messages longer than 127 chars
* Correctly generated a random Sec-WebSocket-Key
* Adds ability for connecting to a websocket on a subdirectory by passing in the path
* Added function to close websocket

Tested working on Espruino WiFi

### DIFF
--- a/modules/ws.js
+++ b/modules/ws.js
@@ -7,8 +7,6 @@
  Websocket implementation on Espruino, it let you control your Espruino from the cloud without the need to know it's IP.
  You will need to use it with a websocket server.
 
- Limitations: The module only accept messages less than 127 character.
-
  How to use the ws module:
 
  ```javascript
@@ -17,6 +15,7 @@
  // =============================== CLIENT
  var WebSocket = require("ws");
  var ws = new WebSocket("HOST",{
+      path: '/echo',
       port: 8080,
       protocolVersion: 13,
       origin: 'Espruino',
@@ -61,6 +60,16 @@
 
 /** Minify String.fromCharCode() call */
 var strChr = String.fromCharCode;
+var crypto = require('crypto');
+
+function buildKey() {
+  var randomString = btoa(Math.random().toString(36).substr(2, 18));
+  var toHash = randomString + '258EAFA5-E914-47DA-95CA-C5AB0DC85B11';
+  return {
+    source: randomString,
+    hashed: btoa(crypto.SHA1(toHash))
+  }
+}
 
 function WebSocket(host, options) {
   this.socket = null;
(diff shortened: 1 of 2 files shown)


AverageMarcus:
Looks like this still has an issue with packets that span two frames.

gfwilliams:
Great, thanks!

I'm a bit iffy about the whole nextFrame thing though? I'm not really sure I understand what's going on (lastData was supposed to handle large packets)... It seems like it would be possible that a normal message could contain character code 129 normally, and then it would cause some problems for the code?

Merging anyway as it's going to be a big improvement :)

AverageMarcus:
Yeah the nextFrame stuff was very much achieved through a lot of trial and error last night. The existing code didn't do any checking for the start of a frame being in the middle of the data it is working on (it only checked the first byte) so this is a way of seeing if there are additional frames and splitting them out to be processed next.

From my testing it seems to work fine. Tested with a packet that had about 5+ frames in the data.

gfwilliams:
Is there a public URL you tested with that I could have a go on?

I imagine that with the masking thing (so numbers in the message are more or less random), if you leave it running it may only be a matter of time before something breaks...

I guess rather than searching for the start of data character, we should just make sure we know how long each specific message type is supposed to be and then handle it and take those bytes off the front of the message before processing the next one? It looks like the old code sort of did that for messages but then it just left the data in `lastData` - it never actually checked it for the next message type.

So if you'd sent:

```
[message one][message two][message t
hr
ee]
```

(one line per call to that handling function)

it would have given you all three messages, but obviously nothing ever works like that in reality :)

AverageMarcus:
For the multiple frame issue I was using Pusher. Take a look at this: https://github.com/AverageMarcus/pusher-websocket-espruino If you subscribe to several channels immediately after one another you then get messages back with multiple frames. 

For other testing I was using echo.websocket.org which just sends a message maching anything you send to it.

gfwilliams:
Hi,

Do you think you could check the version on GitHub now and see if it works for you? I won't make it go live just yet, but I reckon it deals with the edge cases a bit better (like pings that contain data and getting 2.5 messages in one packet).

AverageMarcus:
I'll give it a try when I get home and post some results.

AverageMarcus:
Check here for code and response: https://gist.github.com/AverageMarcus/e39771bf20aa18d43b8d01069c9f219a

gfwilliams:
Thanks - looks good! Where does the response come from? I know I removed that from the `open` event, but I couldn't really figure out how it was supposed to work... Maybe there should be a content-length in the header  that we pay attention to?

AverageMarcus:
Which response? I have just noticed that I left in some code that doesn't do anything now. I'm going to open another PR to clean it up.

AverageMarcus:
Actually, ignore me I just noticed you fixed it.

gfwilliams:
> Which response?

```
Connected undefined
```

in your commit I found:

```
this.emit('open', data.length ? data.substring(data.indexOf('{')) : data);
```

which made me think maybe there was some other data send (but I don't see it?)

P.S. this might make you happy. It turns out it wasn't so hard: https://github.com/espruino/Espruino/commit/58ce4a62bb5eede87efde96b1f3c4e801a3d3b5c

AverageMarcus:
Yeah I left that in by mistake.

That is amazing!!!

PR #293, opened 2016-12-01
Update notes about experimental flag

beaufortfrancois:
Have you had any issue with running BlueZ 5.43 without experimental flag?


gfwilliams:
Great, thanks! I am actually running with Bluez compiled exactly as on that page, and it does work.

Having said that sometimes it'll just stop working randomly... I'll be connected and everything will be great, but suddenly I'll stop getting data back from BLE. I kind of assumed that was one of those many patches you'd had to do for Chromebook that hasn't made it in yet :)

PR #296, opened 2016-12-01
Add missing www to puck-js URL

beaufortfrancois:
It looks like https://puck-js.com certificate is not valid while https://www.puck-js.com is.

```
Certificate Error
There are issues with the site's certificate chain (net::ERR_CERT_COMMON_NAME_INVALID).


gfwilliams:
Thanks! Will have to check that out - let's encrypt should have handled it but it looks like it might be trying to use one certificate for both sites

PR #298, opened 2016-12-04
New module for pressure sensor HP03S

luwar:
This is a software module for a pressure sensor. It's an old one but it is still sold.

The module code can be advanced minified with Google closure compiler. I completed espruino.externs and wrote an HP03S.externs file.


gfwilliams:
Nice - thanks! I guess we should probably have a way to create the externs file automatically eventually - maybe some hack based off espruino.json?

luwar:
The espruino.externs is generated directly from the jswrap.c files but the parser and generator is written in Groovy.  That's why I did not contribute it. That would only add an additional puzzle to the already complicated build steps (IMO).

I'm looking for a more unified solution, e.g. parsing the jswrap files with Python or Node into a real model structure (not only generic maps) and then use a template engine to generate the different artifacts with simple template files. 

I stopped because I was not sure about the prefered build language. Python or Node? Any preferences? I would prefer JavaScript to Python. Python is okay too. But I read something about OpenWrt support without cross-compiling. Is this still valid? Is using Gulp allowed?

Do you have any suggestions? I would like to make a new try.

gfwilliams:
Ahh, right. Yeah, I'd like to stick with just Python + Node.js as much as possible.

Honestly, I'd like to move to Node, but quite a lot of the build process stuff does still use Python - and there's a lot to be said for being able to build for constrained devices like OpenWRT without Node.js.

My current feeling is: Use Node.js where possible *outside the build* - especially for things like documentation - but make sure the build can be done with only Python.

> I'm looking for a more unified solution, e.g. parsing the jswrap files with Python or Node into a real model structure

There's a branch of Espruino called 'real_prototypes' or something that really improves this and produces a much neater set of data, but I never managed to finish it off and it's a bit out of date now.

However coming up with a whole new way of parsing jswrap files (there are already 2, one for JS and one for Python) strikes me as the complete opposite of unifying :)

Probably your current solution (working from the JSON) is best, as if I ever manage to finish that branch the actual jwrap format will change anyway.

PR #302, opened 2016-12-12
Updated Windows Installer link

tricki:
Updated Windows Installer link to point to specific section about native app

see http://forum.espruino.com/conversations/297260/#comment13365279


gfwilliams:
Great, thanks!

PR #310, opened 2016-12-26
Small fix in instructions for booting without loading code

joakimstai:
Just noticed that it blinked 5 times, not 3. Made me question if it had booted correctly..

Also, is this the smallest pull request ever?


gfwilliams:
:) Thanks! Yes, looks like that's just a mistake in the instructions

PR #314, opened 2017-01-06
Fixed Documentation

CanyonCasa:
Set Module as first keyword so automatically added to Espruino website module list.

Gordon, while perusing the modules folder I noticed a number of modules that aren't listed on the website Modules page, including the 3 modules I submitted before the holidays. I'm guessing this is due to "Module" not being the first keyword. I have fixed my files, but a number of other modules need fixed as well, such as Grove, GPS, big_number, ... I'm sure others would be interested in seeing and using this body of work.

Not being real familiar with the website documentation generation I chose not to mess with any other module files. Some of them appear to come from other lists such as devices and peripherals. Given your guidance, I'd be happy to help fix others. I think it would also be helpful to emphasize (in bold or red or something) the note in Writing Modules that Modules must be the first keyword including a reasoning  like "as it's used to automatically generate the modules list." It's a bit subtle to others not familiar with the generation process.


gfwilliams:
Thanks - I've just added the keyword to those ones.

Was there a reason you removed the APPEND_USES stuff? It means that if anyone uploads a tutorial in the future that mentions your modules, the page will automatically link to it... Also, swapping the order of keywords shouldn't make any different either?

CanyonCasa:
> Was there a reason you removed the APPEND_USES stuff?

Over zealous. I restored them.

> Also, swapping the order of keywords shouldn't make any different either?

Then it must be case sensitive? As I had module, not Module, in the keywords list.

gfwilliams:
Ahh - got it. It's not case or order sensitive.

If you look at the other modules they follow a pretty standard pattern:

https://raw.githubusercontent.com/espruino/EspruinoDocs/master/devices/AT25.md
```
<!--- Copyright (c) 2014 Spence Konde. See the file LICENSE for copying permission. -->
SPI EEPROMs (like AT25256)
========================

* KEYWORDS: Module,EEPROM,storage,rom,AT25256,AT25512,AT25128,AT25640,AT25320
```

It's what's suggested in http://www.espruino.com/Writing+Modules and the scripts for the website expects stuff in that format.

You're doing this

```
<!--- Copyright (C) 2016 Enchanted Engineering. See the file LICENSE for use. -->
  		  
# Simple Linux-like Cron system
  		  
* KEYWORDS: Espruino, module, cron, cronjob, task
```

If you change the titles to the way they're done on the other tutorials it should work fine.

CanyonCasa:
Ok - changed the titles. I was playing around learning Markdown and different ways to format headers when I made these. I can see how it could affects scripts for extracting the title.

gfwilliams:
Cool - thanks! That should fix it I reckon :)

It's just me being lazy in the scripts really - but it's nice to keep all the files in basically the same format too.

PR #315, opened 2017-01-09
increase timings for sim800 and change reset approach

angelo0000:
The SIM800 needs the reset pin pulled LOW for 200ms to trigger a hard reset. In looking at the SIM900 doc it is the same.  I'm not sure why the reset logic is how it was with pulling it high for 1ms as that doesn't seem to be the correct way to do it.

I also increased the timings on the timeouts b/c the SIM800 doesn't like the 500ms timings.  Tested a few rounds with 1000ms and it seemed to be ok with those.

### DIFF
--- a/devices/SIM900.js
+++ b/devices/SIM900.js
@@ -212,7 +212,7 @@ var gprsFuncs = {
             return cb;
           } else if(r === 'OK') {
             s = 1;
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! actually I'm pretty sure when I did it I never used the reset pin, so maybe the original author didn't either?

PR #319, opened 2017-01-21
Use HTTPS for embedded video

jjok:
Currently the video doesn't show up in Chrome/Linux as it says it's not secure.

There are probably others that need changing to, but this is the only one I've seen.


gfwilliams:
I guess this when viewing from HTTPS rather than HTTP? Seems to work on HTTP for me, but not HTTPS. I wonder if that's a Chrome M56 thing.

I'm pretty sure this will break the script that links up videos. I'll file a bug for this and will change the script itself - as there are probably 20 or more video links in this repo.

PR #326, opened 2017-02-01
add module BMP280 as subset of BME280

MaBecker:


MaBecker:
Git and I still have bad moments ....

devices/BMP280.js
```diff
@@ -0,0 +1,160 @@
...
+};
+
+/* Calibration of Pressure, algorithm is taken from the datasheet */
+BMP280.prototype.calibration_P = function(adc_P) {
+/* enable this if you want to be able to debug data values and calibration
+  if (this.debug) {
+    console.log("T1: " + this.dig_T1);
+    console.log("T2: " + this.dig_T2);
+    console.log("T3: " + this.dig_T3);
+    console.log("H1: " + this.dig_H1);
```
gfwilliams: Please could you just remove these 6 lines? I think they reference the values that were removed?

Looks great though - thanks!

MaBecker: thanks for checking - sure - see commit below

gfwilliams:
thanks!

MaBecker:
Which file do I have to update to see BMP280 in the list of modules ?

gfwilliams:
You'll have to wait for me to update the website I'm afraid - it's not automatic.

You should be able to just do `require("https://raw.githubusercontent.com/espruino/EspruinoDocs/master/devices/BMP280.js")` for now though

MaBecker:
ok, I understand, no need to hurry 

require( "with url") works fine 

thank

PR #327, opened 2017-02-02
Add SHT2x module

urish:


gfwilliams:
Great - thanks!

luwar:
The SHT2x and the [HTU21D](http://www.espruino.com/HTU21D) sensors have the same software protocol.

It's good to have distinct modules if somebody is searching a suitable Espruino module and does not know this fact. But no separate source code is required under the hood.

How about `exports.connect = require("dht21").connect;` in sht2x.js?

gfwilliams:
Argh, that's annoying. Perhaps we should actually just make the HTU21D page mention that it supports SHT2x too, and then have SHT2x forward to HTU21D - since that seems to have a few more features?

PR #328, opened 2017-02-03
update EspruinoESP8266.md

MaBecker:


gfwilliams:
Thanks for the updates - but please can you leave the paths flat (ie. not `/tutorials`) as that'll break the HTML files that are produced by the build script

MaBecker:
if `/tutorials` is missing the link is broken - like now

gfwilliams:
Only in the `.md` file on GitHub - and many things are broken in that.

On http://www.espruino.com/EspruinoESP8266 it works fine

Also - if you viewed this as a file on a PC, `/tutorials` would also be broken I guess as it'll try looking in the root drive of the computer.

Ideally the script would be changed to remove the complete path from the link, and all the other links in EspruinoDocs would be changed - but I think it'd unlikely anyone is going to do that :)

MaBecker:
Thanks, now i understand

gfwilliams:
Thanks! However it still has the `.md` - same reasons I'm afraid.

Does GitHub have a way of doing banners? I think it'd help if I could have this repo mention that the 'compiled' docs are on the main Espruino site - it's a big help having APPEND_JSDOCS/etc working properly

MaBecker:
Ups, sorry ...

gfwilliams:
Thanks! :)

PR #329, opened 2017-02-03
add iBeacon module

urish:


gfwilliams:
Argh, unbelievable. I literally just did one of these this morning, and the `git pull` `git push` had failed so it didn't update :(

https://github.com/espruino/EspruinoDocs/blob/master/modules/ble_ibeacon.js

and docs: https://github.com/espruino/EspruinoDocs/blob/master/puck/Puck.js%20iBeacon.md

Does your work ok? Mine seems to have another 3 bytes at the start.

I like yours though. Sorry about this. Assuming it does work fine, can I pull the code out of yours and into mine? I'd tweaked the layout a bit so you can use the data directly to do interleaved advertising if needs be

urish:
Yes, it works (checked with the nRF Connect app). Feel free to pull
anything you need from it - it doesn't really matter to me which version
will make it into the code, as long as there will be some implementation.

Uri

gfwilliams:
Sorry - closing this one now and nobody seemed to have issues with the ibeacon I'd done

PR #331, opened 2017-02-07
EspruinoWifi: bugfix for scan function

PaddeK:
Following code
```
var wifi = require("EspruinoWiFi");

wifi.scan(function() {
  console.log(arguments);
});
```

Results in
```
>Uncaught Error: Field or method "cmdReg" does not already exist, and can't create it on undefined
 at line 1 col 11
var b=[];d.cmdReg("AT+CWLAP\r\n",5E3,"+CWLAP:",function(a){a...
          ^
in function "scan" called from line 3 col 2
});
 ^
=undefined
> 
```

This PR fixes that.


gfwilliams:
Ahh - nice, thanks!

PR #336, opened 2017-03-09
Strange heading in About Bluetooth LE

OwenBrotherwood:
Remove TL;DR -
Not too sure if this is correct or if I I do not have knowledge of what TL;DR - could be

Unless this is a RTFM in urban, which is a bit obscure ...
http://www.urbandictionary.com/define.php?term=tl%3Bdr

Just a quick PR for an observed little thing

Signed off
Owen Brotherwood

image

### DIFF
--- a/puck/About Bluetooth LE.md
+++ b/puck/About Bluetooth LE.md
@@ -18,7 +18,7 @@ There are two distinct types of device (Puck.js can be both):
 * **Peripheral** is a device that gets connected to, like a fitness band.
 
 
-TL;DR - Super-quick summary
+Super-quick summary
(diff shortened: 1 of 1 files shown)


gfwilliams:
It's intentional...  https://en.wikipedia.org/wiki/TL;DR

I guess a lot of people won't understand, but many will probably appreciate the geekiness :)

OwenBrotherwood:
Yep: when I found the geekiness, it was cool: I had noticed it a few times and was allways wondering.
It was only after I found the link in urban AFTER PR that I began to geek in/out on it:)

PR #337, opened 2017-03-09
Module for Vizibles IoT platform 

pablorodiz:
Added the code and documentation for a new module for Espruino which helps on connecting things to the new [Vizibles](https://vizibles.com) IoT platform.

### DIFF
--- a/info/IoT Services.md
+++ b/info/IoT Services.md
@@ -10,6 +10,55 @@ There are [quite a lot](http://postscapes.com/companies/iot-cloud-services) of I
 **Note:** To avoid duplication, all of this code expects that you already have an [Internet connection](/Internet).
 
 
+Vizibles
+--------
+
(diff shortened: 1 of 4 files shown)


gfwilliams:
Great, thanks! I'll tweak the URLs to the Spanish version a bit so that they work on the main espruino website (rather than just GitHub).

Where you do a bunch of `at.write(`'s you could just do `at.write(a+b+c+d)` - it'd minify down an awful lot better. Not sure if you want to change that?

Also, the Vizibles.js file looks pretty heavily based on the existing ESP8266 lib? It'd be nice to also give credit to the original authors in the header.

pablorodiz:
Ok, I'll take your suggestions. Please wait for my next commits before merging.

pablorodiz:
Done. And please feel free to tweak the URLs.

gfwilliams:
Great - thanks!

PR #338, opened 2017-03-11
Add WebServer module

EthraZa:
Don't have Apache or Nginx to Espruino? Your problems are over! 
Adding Espruino WebServer v1.0 module.

### DIFF
--- a/modules/WebServer.js
+++ b/modules/WebServer.js
@@ -1,0 +1,226 @@
+/* Copyright (c) 2017 Allan Brazute. See the file LICENSE for copying permission. */
+/*
+ Espruino Web Server
+
+ * KEYWORDS: Module,Espruino,webserver,Internet,http,ajax,html
+
+ Webserver let your Espruino unit act like a real web server and serve static and server generated content.
+
+ How to use the WebServer module:
+
+ ```javascript
+ // StartAP or connect to WiFi, then...
+
+ var WebServer = require("WebServer");
+ var webs = new WebServer({
+		port: 80, // Default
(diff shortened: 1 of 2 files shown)


gfwilliams:
Looks great, thanks! Are you sure about `String.prototype.byteLength` though? since there's no unicode in Espruino the bytelength always equals the normal length.

wilberforce:
@EthraZa

The use of trying literals might simplify of the code... 


```
for (var i=0;i<5;i++) 
  console.log(`No.${i} Hello World, even ${i*0.25}`);

```

wilberforce:
Trying literals - should be string literals !

gfwilliams:
String literals also allow you to have newlines in them, so including webpages is really easy now :)

EthraZa:
Hi, yeah, the byLenght was just a long shoot to try to have a Content-Lenght. Since I know its probably wrong sometimes, maybe it's just better to cut it off to save space.
About String literals, yeah, I know, but they still look so strange to me that I get my self avoiding them... I need a mental work here to force my hand to use it :)

I'll improve that stuff for the next version where I'll try to combine it with @wilberforce [FlashStore](http://forum.espruino.com/conversations/283045/)

Thx

wilberforce:
@EthraZa
After seeing your interest in the forums, I had another look FlashStore. Since it was first made, something has changed with the way prototype works and the way I was adding new methods on the fly doesn't work any more....


If you are playing to use this stuff on the esp8266, You should see how many js vars you have left as I have feeling you will be running out of space

PR #342, opened 2017-03-15
Enhance Vizibles documentation

pablorodiz:
Changed Key Secret and Key ID on Vizibles module examples to make them abviously wrong, so force averyone to create their own instead of testing generic (wrong) but credible values.

### DIFF
--- a/modules/Vizibles.es.md
+++ b/modules/Vizibles.es.md
@@ -47,8 +47,8 @@ connect = function(options, callback, connectionCb, disconnectionCb)
 `options` es un objeto con pares clave/valor para configurar la conexión 
 ```
 {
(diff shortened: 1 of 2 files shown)


gfwilliams:
:) Thanks! Has that been a problem?

pablorodiz:
It was not a big problem, since the key writen in the example is public for creating public things on Vizibles. But the real problem, or at leas for one of our internal testers it was, is if two people try to use the same combination of key and id (what will surely happen if people just copy & paste the example as it was). The second person can not acces or control the thing, but without any error. Since the platform already knows a thing with the same key and id it will be the same thing. And belong to the same owner... That is the problem.

PR #347, opened 2017-04-21
fix typos, re-insert pinout

wilberforce:


gfwilliams:
should this me merged? I missed this.

PR #351, opened 2017-05-07
precision Maker Webhooks (over WeMo Maker)

da4throux:


gfwilliams:
Great, thanks! It seems they've changed things a bit since I wrote this.

PR #352, opened 2017-05-08
Error correction in setMT(mt) function.

Delanet:
The function setMT(mt) did not work. The code was copied from the datasheet BH1750, but it is not working: `1 / 1.2 *( 69 / X )`, where `X` is MTreg value.
The variable `X` caused an error in `this.factor=1.2/(69/X)`.


gfwilliams:
Great - thanks! At some point in the past Espruino didn't error for undefined variables, so I guess this bug probably crept through then

PR #354, opened 2017-05-15
Reconnect to MQTT on timeout

wanglingsong:
Added one more option "RECONNECT_INTERVAL" in constants C

### DIFF
--- a/modules/MQTT.js
+++ b/modules/MQTT.js
@@ -263,7 +263,7 @@ MQTT.prototype.connect = function(client) {
   };
   if (client) { onConnect(); }
   else {
-    var net = require("net")；
+    var net = require("net");
     var options = {host : mqo.server, port: mqo.port};
     client = new net.Socket(options);
     client.setTimeout(mqo.C.CONNECT_TIMEOUT, function() {
--- a/modules/MQTT.js
+++ b/modules/MQTT.js
@@ -64,7 +64,8 @@ function MQTT(server, options) {
 MQTT.prototype.C = {
   DEF_QOS         : 0,    // Default QOS level
   CONNECT_TIMEOUT : 5000, // Time (s) to wait for CONNACK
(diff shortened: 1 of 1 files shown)


wanglingsong:
btw, the module minification of web IDE hardly work. So I have to manually add changed code into your minified source for testing. That was terrible!

gfwilliams:
Thanks! I'm not sure what you mean by the minification not working - do you have any example? Could it be that you were using the online minifier and you had a bad internet connection?

wanglingsong:
What minifier and minifier config used for the official min module? I didn't find it in the document.

gfwilliams:
It's the closure compiler - you can find how it's done in `bin/minify.js`

wanglingsong:
Thanks!

jjok:
@gfwilliams This seems like a useful feature. Do you know why it was reverted in the next commit?

https://github.com/espruino/EspruinoDocs/commit/46f1054d843df9ed77aecbf1a5d8a77fec51ff64#diff-d136e58fdb064ae71881628760526607

Was that intentional, or did someone forget to rebase before that massive commit?

gfwilliams:
It probably was someone who forgot to pull latest before their commit :( Some people start by massively refactoring everything to the way they like it, and while I try not to merge stuff like that if loads of people are reporting issues that were fixed in their changes then eventually I end up caving in and stuff like this happens.

Having said that I believe there are a few examples of auto-reconnect and it's only a few lines to do manually from the `disconnected` event.

It's important that we *don't* automatically default to connecting using `net` though as some people want to use `tls` to connect, and I recall there are even people using non-sockets (so MQTT over Serial, Bluetooth or other stuff)

jjok:
OK, thanks. I saw the PR that added the feature and thought that I might be able to replace my own reconnect code with that, then was confused that it didn't exist anymore. No worries.

PR #355, opened 2017-05-16
Add TLE94112 Module

JoKl70:
This will add support for Infineon TLE94112 Motordriver. Currently the implementation supports to control 6 DC Motors and reading status registers. Support for LED drivers will be added later.


gfwilliams:
That's great - thanks!

PR #357, opened 2017-05-21
LIS2DH12: Fixed function name.

drandreas:


gfwilliams:
Argh - thanks for pointing that out!

PR #360, opened 2017-05-25
Documentation for ESP32 tweaks

tomgidden:
Just a few doc tweaks: some mistakes, whitespace issues. Also, `wifi.getIP()` ⇒ `wifi.getIP().ip`.  Feel free to cherry-pick.

Incidentally, where does a binary user get hold of `bootloader.bin` and `partitions_espruino.bin`?  The release package only contains `espruino_1v92_esp32.bin` if I remember correctly. I had to do the source build for my initial flash.

Tom

### DIFF
--- a/boards/ESP32.md
+++ b/boards/ESP32.md
@@ -259,8 +259,8 @@ var password = 'YOUR_SSID_PASSWORD';
 
 var wifi = require('Wifi');
 wifi.connect(ssid, {password: password}, function() {
-    console.log('Connected to Wifi.  IP address is:', wifi.getIP());
-	wifi.save(); // Next reboot will auto-connect
+    console.log('Connected to Wifi.  IP address is:', wifi.getIP().ip);
+    wifi.save(); // Next reboot will auto-connect
 });
 ```
 
@@ -295,7 +295,7 @@ wifi.connect(ssid, {password: password}, function() {
         res.end('Hello World');
     }).listen(port);
 
-    console.log(`Web server running at http://${wifi.getIP()}:${port}`)
+    console.log(`Web server running at http://${wifi.getIP().ip}:${port}`)
 });
 ```
 
--- a/boards/ESP32.md
+++ b/boards/ESP32.md
@@ -254,8 +254,8 @@ If your board with the ESP32 has an antenna connection point, connect your Wifi 
 boards have an in-built antenna so this may not be required.
 
 ```JavaScript
-var ssid = 'YOUR_SSID';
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! Might be an idea to replace `const` -> `var` though? `const` is part of ES6 and doesn't actually work properly in Espruino (it's just treated as `var` at the moment because I had to do something with it)

I think there's a PR that's been merged so the next release will have all the extra bin files in a tar. For now I think you probably need to drag them out of GitHub: 

https://github.com/espruino/EspruinoBuildTools/blob/master/esp32/deploy/app.tgz
https://github.com/espruino/EspruinoBuildTools/pull/6/files

tomgidden:
Oh, okay.. I noticed that `port` was a `const`, so I just changed the `var`s to `const`s for consistency.

Is it worth putting in a note about those files in this text file, or is the next release imminent (enough)?

wilberforce:
Thanks.

The bin files are now part of the travis builds:
https://www.espruino.com/binaries/travis/master/espruino_1v92.64_esp32.tgz

So the next release 1v93 would have them packaged too.

wilberforce:
The other thing that is not really apparent is how to get hold of the flashing tool v2 to flash the firmware without downloading the espruino.git

tomgidden:
I've added a couple of changes reflecting @wilberforce and @gfwilliams' comments.  No solution to the flashing tool, mind you.  I seem to pick up various copies of `esptool.py` from all sorts of installations, but I don't think I've ever explicitly asked for it...

gfwilliams:
Great, thanks! Yes, `const` does work its way into example code - I just don't want to encourage it since it'll cause someone problems at some point. Same with `let`.

Thanks for the tweaks! I think 1v93 will be a few weeks out, but I'm hoping it won't be as long coming as 1v92 was!

boards/ESP32.md
```diff
@@ -146,6 +147,14 @@ esp-idf/components/esptool_py/esptool/esptool.py    \
         0x10000 espruino_esp32.bin
 ```
 
+If the `bootloader.bin` and `partitions_espruino.bin` files are not
+included from the same source you acquired the `espruino_esp32.bin` build,
+they should be available in the corresponding `espruino_1v92.*_esp32.tgz`
```
tomgidden: Oops... should be `espruino_*_esp32.tgz`.  Anyway, it sounds like this addition will be imminently obsolete by the time it matters.  It could be removed then.

PR #368, opened 2017-06-17
Jo kl70 Add VCNL4020 support

JoKl70:
Add support for VCNL4020 Integrated Proximity and Ambient Light Sensor with
Infrared Emitter, I2C module.
The module implements the API as recommended by Vishay.

### DIFF
--- a/devices/VCNL4020.js
+++ b/devices/VCNL4020.js
@@ -1,0 +1,203 @@
+/* Copyright (C) 2017 Joachim Klein. See the file LICENSE for copying permission. */
+/*
+This module interfaces with Vishay VCNL4020. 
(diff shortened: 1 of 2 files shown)


gfwilliams:
This is awesome - thanks! I'll get this on the main site in the next day or so.

Just to add - because by default the modules are minified, the minifier is smart enough to rip all the `DEF`s that aren't used out (as long as `DEF` itself isn't exported), so you don't have to worry too much about stripping elements out if you don't want to.

JoKl70:
Hi Gordon,

thank you for the feedback. I will considere this for future modules.

Best regards

Joachim


Am 19.06.2017 um 13:57 schrieb Gordon Williams:

PR #373, opened 2017-07-23
VL53 Module, add support for more than one sensor on the same bus

JoKl70:
This change will add support for the changeAddress feature of the VL53. By using this feature it's possible to use more than one sensor on the same I2C bus.

### DIFF
--- a/devices/VL53L0X.js
+++ b/devices/VL53L0X.js
@@ -4,11 +4,19 @@ var C = {
   REG_RESULT_RANGE_STATUS : 0x0014
 };
 
+/** 
+Init VL53 Sensor with default address.
+Store Address in a variable, this allows to change the address later.
+*/
 function VL53L0X(i2c, options) {
   this.options = options||{};
   this.i2c = i2c;
-  //this.r(0xC0,1)[0]==0xEE;
-  // initialise
+  this.ad = 0x52>>1;
+  this.init();
+}
+
+/** initialise VL53L0X */
+VL53L0X.prototype.init = function() {
   this.w(0x80, 0x01);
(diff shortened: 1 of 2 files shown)


gfwilliams:
Great, thanks! Do you think it makes sense to allow the address to be set in `options` when the module is initialised the first time?

JoKl70:
Yes, this might make sense. 
I'm not sure how to correctly implement it. Do you have an example how to use options?
To be compatible with the previous implementation I don't like to use an additional parameter.
The implmenentation is currently similar to other implementation, e. g. for Arduino.

JoKl70:
Sorry, fist time I use this feature. 
Thought Close and comment will story and forward the comment.

gfwilliams:
I'd maybe do something like:

```
function VL53L0X(i2c, options) {
    this.options = options||{};
    this.i2c = i2c;
    this.ad = 0x52>>1;
    if (this.options.address) {
     // Change I2C address, if specified in options
     this.ad = this.options.address>>1;
     this.i2c.writeTo(0x52>>1, 0x8a, this.ad);
    }
    this.init();
}
```

Then you can instantiate with `VL53L0X(i2c, {address:0x56})`. In that case you wouldn't even need the changeAddress function, since it's not that much use (you could only ever call it once, at initialisation time).

What do you think?

JoKl70:
Hi, got it.
I changed the call for VL53L0X as recommended. Now it's much shorter.
The descitpion is also changed to new version.
Tested with STM32F411 Discovery and tweo VL53L0X.
Thank you.

gfwilliams:
This looks great - thanks!

Merging now - it should be online in a few days

PR #376, opened 2017-08-17
InfluxDB module

Sparow199:
ADD InfluxDB module with documentation.


gfwilliams:
That's great - thanks! I'll get that pulled in.

Just FYI - the title right at the top needs to be the standard Markdown style because the scripts in this repo do some sanity checking on the input - but I'll tweak it so it works.

PR #378, opened 2017-08-18
UG96 module introduction, a QUECTEL modem in Espruino

tbistm:
This is a library inspired from SIM900 and apllicable for UG96 from QUECTEL.
Similar functionnalities are available :
- full modem initialization,
- modem reset,
- debug with 4 levels,
- socket management, up to 12 in parallel,
- TCP client working in Direct Push Mode
- Flow control to deal with large HTTP frame (until 1460 bytes) and
  throughput from different cellular technologies.

This is done with the AT commands set from QUECTEL.

New Positioning API is also implemented
It gives your position from the cellular.

Several HTTPs (upload & download) done on an evaluation board.

### DIFF
--- a/devices/UG96.js
+++ b/devices/UG96.js
@@ -1,0 +1,1036 @@
+/*
+  Copyright (c) 2017, STMicroelectronics - All Rights Reserved
+  Largely inspired from SIM900 modem library:
+  Copyright (c) 2015 Gordon Williams, Tobias Schwalm.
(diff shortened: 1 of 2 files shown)


tbistm:
Hi,
For further details, you can contact me by e-mail : thierry.biniguer@st.com
Regards.

gfwilliams:
Thanks! Looks awesome!

PR #381, opened 2017-09-10
Some improvements for the ESP32 document

edwinm:


boards/ESP32.md
```diff
@@ -212,7 +229,7 @@ be quite reliable.
 #### minicom / CuteCom
 
 There are other methods to connect to your ESP32, not just Espruino Web IDE.  Two
-very basic tools are [minicom](https://en.wikipedia.org/wiki/Minicom),
+very basic tools are [CoolTerm](http://freeware.the-meiers.org/),
```
gfwilliams: Please can you leave this as minicom, or at least add CoolTerm as another option? Minicom is probably the easiest solution for those running linux

gfwilliams:
Thanks! Merging this - it would have been nice to keep minicom mentioned, but I guess those running Linux will figure it out.

PR #387, opened 2017-10-25
update GPIO16

MaBecker:


gfwilliams:
Thanks! Just to add, i'd have thought the way it's implemented would mean that *software* I2C and SPI would work fine? Also I think the only PWM available atm is software, so that should be fine too.

PR #389, opened 2017-11-03
Added 20ms Timeout for setWatch method

lanselambor:
Low level state last for 20ms is needed before signal pin turn to input mode

### DIFF
--- a/devices/DHT22.js
+++ b/devices/DHT22.js
@@ -20,10 +20,11 @@ DHT22.prototype.read = function (cb, n) {
   var ht = this;
   pinMode(ht.pin); // set pin state to automatic
   digitalWrite(ht.pin, 0);
-  this.watch = setWatch(function(t) {
-    d+=0|(t.time-t.lastTime>0.00005);
-  }, ht.pin, {edge:'falling',repeat:true} );
-  setTimeout(function() {pinMode(ht.pin,'input_pullup');},1);
+  setTimeout(function() {
+    this.watch = setWatch(function(t) {
+      d+=0|(t.time-t.lastTime>0.00005);
+    }, ht.pin, {edge:'falling',repeat:true} );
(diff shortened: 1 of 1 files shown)


gfwilliams:
Did it not work previously? DHT11/22 modules have been through a few different iterations and have been tested pretty extensively on various different platforms and seemed to work pretty well.

Adding a 20ms timeout pretty much guarantees that the setwatch won't be in place until some time after 20ms because of JS execution time, and if the device is busy it could actually be significantly after that, making readings unreliable.

lanselambor:
The present DHT11.js module can not work on the [Grove - Temperature & Humidity Sensor](https://www.seeedstudio.com/Grove-Temperature-%26-Humidity-Sensor-p-745.html), and as the [data sheet](https://akizukidenshi.com/download/ds/aosong/DHT11.pdf) says, a **Host Send Start Signal**   should longer than 18ms, I had tested it that the available Start Signal is between 18 ~ 25ms. 

Can you tell me the sensor your team used, I was wondering if there's another sensor module like DHT11, if I was using a different one, I hope to add a new javascript module for my sensor.

lanselambor:
Ok, things finally clear. There are three types of DHT sensor DHT11, DHT21(AM2301), DHT22(AM2302), the Host Send Start Signal required on DHT11 is longer than 18ms, on DHT21 and DHT22 is longer than 1ms. Refer to this [document](https://kropochev.com/downloads/humidity/AM2301.pdf).

I think the best solution is to set Host Send Start Signal longer than 18ms. 
please check it.

lanselambor:
Please close this issue, compared DHT11, DHT22 and AM230x, only DHT11 need start signal longer than 18ms. I has turned to DHT11.js.

gfwilliams:
Ok thanks!

PR #390, opened 2017-11-03
Added setTimeout for setWatch

lanselambor:
Low level state must last for 20ms before signal comes out. And setWatch can set gpio to input mode.


gfwilliams:
closing as duplicate

PR #391, opened 2017-11-05
correction of wrong register address

JoKl70:
During extended test I figured  out that one address was wrong.


gfwilliams:
Great! Thanks for finding this!

PR #395, opened 2017-11-14
Set Start Signal stay low for 18ms

lanselambor:


gfwilliams:
Has this been tested? You commented out the `pinMode` function, so as far as I can tell the pulse never ends!

Surely all you need to do is change `setTimeout(function() {pinMode(ht.pin,'input_pullup');},1);` to `setTimeout(function() {pinMode(ht.pin,'input_pullup');},20);` and leave the timeout for setwatch.

I have a DHT11 here now so I'll test and will get back to you

gfwilliams:
Ahh, ok - sorry. I did some testing, and `setWatch` sets the pin state to input automatically - which I didn't understand was the problem initially.

I'll tweak this and the DHT22 to work around it.

PR #396, opened 2017-11-20
Ug96 upgrade

tbistm:
New features have been developed and can be used from the application :
- power on/off the modem (API to start/stop the modem for the power control)
- set the network configuration (APN) based on the SIM identity (CCID)
Improvements are also proposed to optimize time to registration and socket management


tbistm:
Hi, could you have a look at this PR ? Thanks.

gfwilliams:
Thanks! Sorry, this got lost under all my emails.

PR #397, opened 2017-11-21
Fixed Error in Docs. You need to write to UART TX, not RX

ThomasChr:
I'm pretty sure I've spotted an error here :-)


gfwilliams:
Great - thanks!

PR #399, opened 2017-11-30
Update Chrome platform support for Web Bluetooth

scheib:


gfwilliams:
Hi - thanks! 

Isn't: `Linux support in Chrome is not yet available` a bit confusing? I mean, there are a bunch of people using it and it actually says there's support on the implementation website: https://github.com/WebBluetoothCG/web-bluetooth/blob/gh-pages/implementation-status.md

Also, does Android 5 really not work any more? Amazingly there are still people out there with Android 5 devices who care about it (that bit had to be added because of them!).

scheib:
Now that we've actually shipped on several platforms and support them, and no chrome developer is paying attention to the unsupported platforms, I think their status should be clearly indicated as not supported.

I've reworded Windows, Linux and Android sections.

Linux is not supported.  People can try to use it... but if it doesn't work and folks file bugs they won't be worked on.  The implementation site says 

> Linux & Windows are partially implemented, but not supported. The chrome://flags/#enable-experimental-web-platform-features flag must be enabled.

Android 5 is similarly something that may break at any time, though I admit it's less brittle than Linux.  But, no one working on Chrome is checking/using it.

gfwilliams:
Looks great - thanks! Yes, the Linux support is definitely a bit flaky - I have to restart `bluetoothd` pretty often.

PR #410, opened 2018-01-23
add NFC Tag Emulator

drandreas:
I've hopefully come up with a documentation that consist of enough pointers to enable a user to continue from here. I'm not sure if `APPEND_JSDOC` renders correctly.
Is there a way to verify the output?


gfwilliams:
Awesome, thanks! I'll check it out.

You'd need to run `./build.sh` in the repo and see what it sticks in the HTML folder, but I'll make sure it's fine here

gfwilliams:
Looks spot on - I'll get that on the website this week.

PR #415, opened 2018-01-27
Add Support for STM LIS3MDL Sensor

JoKl70:
This implementation supports I2C.
Tested with STM32 Nucleo and X-Nucleo-IKS01A01.


gfwilliams:
Neat - thanks!

PR #416, opened 2018-02-01
Change h5 to h4

jjok:
There are some `h5` elements, which are currently unstyled, making them difficult to read. I think they're meant to be `h4`,


gfwilliams:
Neat - thanks!

PR #423, opened 2018-03-30
Enable the radio systematically (modem full functionality)

tbistm:
Add a RADIO ON state during modem initialization.
Mandatory to have for BG96 out of its blister (first usage), because radio is unset by default.
Of course, it is applicable for UG96 and nice to have (radio is already set by default).


gfwilliams:
Neat, thanks!

PR #425, opened 2018-04-15
OpenWeatherMap Example

JoKl70:
Hi Gordon,

as discussed in Heidelberg, here comes my open Weather Map Example.
I think this will allow many new demos with Espruino (Weather Station, Lighting Control depending on sunrise / sunset time and many more).

Best regards
Joachim


gfwilliams:
That's great, thanks! I *think* for example code like this it's possible to include the markdown right at the top of the JS file like https://github.com/espruino/EspruinoDocs/blob/master/examples/simple_data_logger.js (to avoid having two different files) - but I'll quickly do that here.

If you have or come up with more examples like this it'd be great to have them in. This would be really useful for loads of people.

PR #432, opened 2018-05-02
Added new device SI7021 I2C temperature and humidity sensor

ukaoskid:
Added a new device interface for SI7021 Silabs I2C Temperature and Humidity Sensor


gfwilliams:
Great, thanks!

PR #435, opened 2018-05-20
Faster TM1640 flip using shiftOut(clkPinsClk)

opichals:
Only to be merged if https://github.com/espruino/Espruino/pull/1431 gets in.


opichals:
Obsoleted by https://github.com/espruino/EspruinoDocs/pull/436

PR #436, opened 2018-05-28
Faster TM1640 flip using sw SPI

opichals:
Using the following setup the pin signaling is the same as was in the original send method:

```
var spi = new SPI();
spi.setup({mosi:pins.din, sck:pins.clk, mode:3, order:'lsb'});
var send = spi.write.bind(spi);
```


gfwilliams:
Thanks!

PR #444, opened 2018-07-17
fix: typo in getGas() docs

urish:
It should read `eCO2` instead of `EC02` (The letter capital O instead of a Zero)


gfwilliams:
Thanks!

PR #448, opened 2018-08-28
Fix GPIOTE samples in NRF52LL

urish:
The GPIOTE functions expect a chip pin number which does not always correspond to Espruino pin number. Updated the samples to reflect that and use `pin.getInfo().num`

### DIFF
--- a/devices/NRF52LL.md
+++ b/devices/NRF52LL.md
@@ -32,7 +32,7 @@ Count the number of times the BTN pin changes state (GPIO + counter timer):
 ```
 var ll = require("NRF52LL");
 // Source of events - the button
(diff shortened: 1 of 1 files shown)


gfwilliams:
Ahh - thanks! I think in other areas of NRF52LL I use `getInfo().num`? Maybe to make things more sane I should just add that to `GPIOTE`?

urish:
You use `getInfo().channel`, which is quite similar.

If you do decide to go with changing NRF55LL, then I suggest to new `new Pin(p).getInfo().num`, so that the user will still be able to pass a pin number and not just a Pin object (same as `digitalRead`, `digitalWrite`, etc.)

gfwilliams:
Thanks!

PR #449, opened 2018-08-28
add NRF52Stepper module + docs

urish:
Enables accurate control of Stepper Motors using nRF52 PPI Hardware


gfwilliams:
Looks great - thanks!

PR #457, opened 2018-09-30
Added a first reference to the inline c compiler

CWBudde:
So far a reference to the inline c compiler was missing. Even worse it was mentioned that it is not yet possible (while it is). This has been fixed to a degree that at least a reference is presented.


gfwilliams:
Thanks! Just FYI but `bin/espruino.json` is computer-generated, so it' should get updated aropund about the next Espruino release.

PR #458, opened 2018-10-03
Create ble_hid_mouse.js

derekarthur:
Bluetooth HID mouse


gfwilliams:
Looks great - thanks! It may also be that for things like movement, the `send` call actually doesn't need to send two HID commands, and just one will do?

PR #465, opened 2018-11-10
Add files via upload - Color module - Step 2 of 4

sleuthware:
Add files via upload - Neopixel project 2018.11.10 Color module - Step 2 of 4

Contains JSDoc comments

### DIFF
--- a/modules/Color.js
+++ b/modules/Color.js
@@ -1,0 +1,270 @@
+/* Copyright (c) 2018 Robin G Cox  See the file LICENSE for copying permission */
+
+/* Work in progress - to avoid change duplication - contact author first */
+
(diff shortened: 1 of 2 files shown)


gfwilliams:
I notice you added:

```
/* Work in progress - to avoid change duplication - contact author first */
```

I'll leave it in, but honestly this is the point of Git/GitHub - people should be able to make changes at the same time as you without breaking things, unless they rewrite *exactly* the same bit of code.

gfwilliams:
I notice you added:

```
/* Work in progress - to avoid change duplication - contact author first */
```

I'll leave it in, but honestly this is the point of Git/GitHub - people should be able to make changes at the same time as you without breaking things, unless they rewrite *exactly* the same bit of code.

PR #466, opened 2018-11-10
Add files via upload - NeopixelCore - Step 3 of 4

sleuthware:
Add files via upload - Neopixel project 2018.11.10 NeopixelCore module - Step 3 of 4

Contains JSDoc comments


gfwilliams:
I mean it's unique to you because nothing else in the EspruinoDocs repo uses it.

I'm not using JSDoc - the code that strips the comments off and creates documentation is unique to Espruino. `APPEND_JSDOCS` is named that because JS DOCS just makes sense as a name.

Maybe at some point in the future I'll switch to using JSDoc, but right now I'm not sure it's a good use of my time to replace something that works just fine for existing modules.

PR #467, opened 2018-11-10
Add files via upload - Quick Start - Step 4 of 4

sleuthware:
Add files via upload - Neopixel project 2018.11.10 neopixelQuickStart code file - Step 4 of 4

No JSDoc comments

It would be nice if a separate folder 'Neopixel' could be created inside 'tutorials' to hold neopixelQuickStart.js
and future tutorial uploads, est five as of Sat 2018.11.10

### DIFF
--- a/tutorials/neopixelQuickStart.js
+++ b/tutorials/neopixelQuickStart.js
@@ -1,0 +1,317 @@
+/*neopixelQuickStart.js
+  Sun 2018.10.28
+  Author Robin Cox 
+
+
+  Quick Start for NeopixelCore library using an Espruino Pico and CJMCU-2812-8 Neopixel strip
+
+  
+  Requires at a minimum 1v99 - System module require("neopixel") - hdwr SPI MOSI pin(s)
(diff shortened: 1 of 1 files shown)


gfwilliams:
It ok if I just format this code as markdown and stick it inside `NeopixelCore.md` ? It seems to be what's needed really - just adding a JS file won't actually translate to anything viewable on the Espruino website, and most people looking at NeopixelCore won't notice your code.

sleuthware:
As you have a tremendous amount of experience and vision here, and you're already set on an idea, I'll let you be the best judge. I'm okay with the insertion into NeopixelCore.md, but need to resolve how to create a numbered part tutorial set.

Have plans (throughout winter) to create a 1 to est 5 part tutorial with NeopixelCore to include palettes and animated effects. Which brings up the question, how to differentiate between those five parts? (each a complete page as opposed to one long re-edited run-on) 

Trying to minimize the number of potential changes by just creating whole files to include here.

PR #470, opened 2018-11-30
Update GPS to use checksum

wybiral:
This merge would update the GPS module to validate each line against the checksum according to NMEA and reject all lines that fail. The purpose is to reduce the chance of returning corrupt coordinates.


gfwilliams:
Please could you check `require("https://raw.githubusercontent.com/espruino/EspruinoDocs/master/devices/GPS.js")` and see if that still works ok for you? Should be identical, but will also create a `line` event.

wybiral:
Yep, everything is still working for me. 👍

gfwilliams:
Thanks! I'll make sure that gets pushed. Nice video by the way :)

PR #471, opened 2018-12-16
Add SNTP module

opichals:
Fixes https://github.com/espruino/Espruino/issues/624


gfwilliams:
That's awesome - thanks!

Issue #473, opened 2019-01-10
Improving the BME680 module


AkosLukacs:
Hi!

Thanks for the help in the forums, I started to look at the BME680 module, if I can reduce it's minified size.
I made some progress: wrapped the whole thing in an IIFE, tried to the constants out from the `C` object, but looks like Esprima doesn't eliminate unused ones, they stay there with shortened name. So just commented out the unused constant for now...

Current / original: 9954 bytes minified, memory usage: 774
Modified: 7633 bytes minified, memory usage: 680
Unused constants commented out: 6594 bytes minified, memory usage: 606
Memory usage is on the Pixl.JS with minimal init code

Fixed some bugs: the module is accidentally modifying / leaking a couple of globals: `data`, `factor`, `heater_res_x100`.

Why the sparse arrays for `par_t`, `par_p`, `par_h`, `par_gh`? Looking at the code, `par_t[0]` and others are never accessed? Remnants from some original code?
```javascript
this.cal.par_gh = [,vc.getInt8(GH1_REG)
                   ,vc.getInt16(GH2_LSB_REG, true)
                   ,vc.getInt8(GH3_REG)];
```
So cal.par_gh is `[ undefined, -20, -14548, 18 ]`.

Current progress is [here](https://github.com/espruino/EspruinoDocs/compare/master...AkosLukacs:patch-2#diff-6c0c9fa98ad3e4d1f78f4854d7d20310) sorry, the diff is a mess :/

gfwilliams:
Hi,

Thanks for taking a look at this. It's great to fix those accidental leaks - any chance they could be pulled out into a PR of their own before all the size changes get made?

There's something strange going on with the constants. Usually, that pattern (private `C` variable) works great. For example http://www.espruino.com/modules/SSD1306.js uses it and minifies to http://www.espruino.com/modules/SSD1306.min.js without issues. I'd really prefer to figure out why it's not minifying rather than swapping everything back to global variables, which seems like a bit of a step back. If it could be made to properly minify `C` then instantly loads of space would get saved.

If it helps, I actually use Closure for minification when I minify the modules for publication, and that'll be doing a much better job.

It seems like `par_gh` itself is accessed, just not `0`. Basically a 3 element sparse array is more efficient than defining a 3 element typed array which is why it's done. The element 0 is undefined so shouldn't actually be using much space - however yeah, it's a hangover from the BME680 code it's based on. Shifting everything down by 1 would definitely be an improvement.

[...]

PR #474, opened 2019-01-10
Global hunting #473 wip

AkosLukacs:
Fixed the globals.


gfwilliams:
Great - thanks! You mention 'wip' - do you want this merged now, or after a few more changes?

AkosLukacs:
This can be merged. I guess the reducing minified size part will be more involved...

gfwilliams:
Thanks! Honestly it may not be that involved - there may just be one statement that is stopping `C` from being collapsed down, and once that is fixed suddenly the size will drop drastically.

PR #479, opened 2019-01-14
fix getVal correctly

SpenceKonde:
What needs to be tested is whether the pot given is valid. Looks like a copy/paste error from one of the other methods.


gfwilliams:
Thanks!

PR #486, opened 2019-01-24
Allow creation of a common anode LED

bartmichu:
This change adds a new parameter to the constructor function which allows to create a common anode LED.


gfwilliams:
Thanks - I guess this could have been done with `invert(true)` before? It just wasn't documented,

I'll get this updated.

bartmichu:
That's correct, you would create cathode LED and then "invert" it.

pon., 28.01.2019, 11:04: Gordon Williams <notifications@github.com>
napisał(a):

PR #488, opened 2019-01-28
Replace Math.clip with E.clip

bartmichu:
Just a little bit of cleaning up. Replace deprecated `Math.clip()` with `E.clip()`.

I'm assuming here that they are 100% compatible and I didn't properly test it because I don't have required hardware.


gfwilliams:
That's fantastic, thanks! Yes, they are identical so swapping is no problem.

gfwilliams:
That's fantastic, thanks! Yes, they are identical so swapping is no problem.

PR #490, opened 2019-02-09
SH1107 module

comp500:
I managed to get a SH1107 display (from the [M5Stick](https://docs.m5stack.com/#/en/core/m5stick)) to work with Espruino. The module is exactly the same as the SH1106, but with a couple of modifications to the initialisation commands.

This hasn't been tested with 128x128 or 128x32 displays, but the original module had code for 128x32 displays so it might work.


gfwilliams:
Nice - thanks!

PR #492, opened 2019-02-12
enhance SH1106 (fix #491)

### LINKED ISSUE #491: Add on() and off() functions to SH1106 as in SSD1306
Now using SH1106 for the first time and missing on() and off().

Just copied  from the SSD1306 module as the codes for on and off are the same.

MaBecker:
add g.on() and  g.off()


gfwilliams:
Thanks! It's such a shame that this is basically two separate modules in one though - I guess I might have to tweak that at some point.

PR #496, opened 2019-03-17
New RcSwitch module for espurino. 

xargon180:
This is a port from the original arduiono c++ library https://github.com/sui77/rc-switch/.

The read feature is currently not implemented.


gfwilliams:
Really nice - thanks!

PR #497, opened 2019-03-19
update EspruinoDocs for tinyMQTT repo changes (docs and module)

olliephillips:
This brings tinyMQTT.md and tinyMQTT.js back into line with the tinyMQTT repo. Sorry for delay, there were a couple of issues opened, which I wanted to get boxed off before updated EspruinoDocs. 

I've followed the markdown doc and pasted in any amendments and additions, so it's not a straight copy of the repo README.  I've not added the min.js file, there were none in there, so I assumed that will be done during the next build/update.

Module got some heavyweight testing and is in really good shape I think. Thanks to @AkosLukacs and @MaBecker for their work and contributions.

### DIFF
--- a/modules/tinyMQTT.js
+++ b/modules/tinyMQTT.js
@@ -2,104 +2,124 @@
 /* Stripped out MQTT module that does basic PUBSUB.
    Originally from https://github.com/olliephillips/tinyMQTT */
 
-var _q;
-function TMQ(server, optns){
-	var opts = optns || {};
-	this.svr = server;
-	this.prt = opts.port || 1883;
-	this.ka = opts.keep_alive || 60;
-	this.usr = opts.username;
-	this.pwd = opts.password;
-	this.cn = false;
-	this.ri = opts.reconnect_interval || 2000;
-	_q = this;
-};
+   (function(){
 
-var sFCC = String.fromCharCode;
-
-function onDat(data) {
-	var cmd = data.charCodeAt(0);
-	if((cmd >> 4) === 3) {
-		var var_len = data.charCodeAt(2) << 8 | data.charCodeAt(3);
-		var msg = {
-			topic: data.substr(4, var_len),
-			message: data.substr(4+var_len, (data.charCodeAt(1))-var_len)
-		};
-		_q.emit('message', msg);
-	}
-};
-
-function mqStr(str) {
-	return sFCC(str.length >> 8, str.length&255) + str;
-};
-
-function mqPkt(cmd, variable, payload) {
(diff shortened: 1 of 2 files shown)


gfwilliams:
Do you have any background as to why everything is now wrapped in `(function(){`? The minfication that's done for the Espruino site actually does the wrapping automatically (and removes it at the end) as part of the minification process.

Adding it seems like a big step back to me and will increase size - no other modules do it so I'm wondering why it's needed in this case. 

Also @MaBecker I tried to explain on the issue you filed https://github.com/espruino/Espruino/issues/1583 why the use of `/x00` in modules isn't a problem with modules or Espruino and why you just have to make sure you escape stuff properly when you copy/paste it into a String. It's a bit frustrating to see you still pushing changes into modules regardless.

olliephillips:
The anonymous function wrapper I added, originally I thought to limit the scope of the `var _q;` declaration, I wasn't sure if it was needed  myself TBH - but then noticed it actually improved minification results when using the closure compiler (I use the online version via the webpage). I don't know why this is the case?!? Others have reported it does not help when using other minifiers, but since I'm the one who creates the `tinyMQTT.min.js` manually after most user contributions I left it in. It is easily removed however, and I can do so if it will impact EspruinoDocs builds?

Re the storage piece. I'm not familiar with all the background to that, only that the previous attempt at this broke the tinyMQTT module and I had to reverse it. @MaBecker reports that using the `String.fromCharCode()` approach allows him to load the module from storage "as is", and it doesn't break the module when loaded directly or over http (when I test), so I saw no harm in including on this basis.

gfwilliams:
Thanks - if you could remove the wrapper that'd be great. If you include it in the code that goes to Espruino I believe it will store an extra scope for each function which will hurt memory usage a lot.

In the code I use for minification I add the wrapper, minify, but then remove the wrapper afterwards: https://github.com/espruino/EspruinoDocs/blob/master/bin/minify.js#L119

What it's doing is telling the minifier that nothing outside can 'see' any of the defined functions except what's in `exports`, so their names can be changed/minified... But once that's done you can remove the wrapper and it's fine.

I don't think those Storage changes make a huge difference either, but the issue is really that the whole 'need' for them comes from a misunderstanding about what's going on (I think), and it's not actually fixing the underlying problem. Anyway, I'll try and explain a bit better in https://github.com/espruino/Espruino/issues/1583 rather than polluting this.

olliephillips:
No problem, I've added a commit to remove that wrapper now.

gfwilliams:
Thanks!

PR #498, opened 2019-03-20
Fixed MQ135.js resistance calculation

timotr:
Resistance calculation used a constant from Arduino which analogRead returns a value between 0 and 1023 whereas Espruino returns between 0 and 1.


gfwilliams:
Brilliant - thanks! Sorry about this - it's really hard to see if some of the contributed modules have been tested properly :(

PR #499, opened 2019-03-30
remove special char

MaBecker:


MaBecker:
JSHint was complaining about special chars ;-)

gfwilliams:
Strange... What did you change? I guess it could be a line formatting char that the forum tends to put into stuff?

MaBecker:
Yes, it was displayed as as red dot 🔴

PR #500, opened 2019-04-10
Smart meter slightly broken formatting

AkosLukacs:
The sample after "Create a new counter. Contains the following:" is broken.
Looks like this:
image


gfwilliams:
Thanks!

PR #501, opened 2019-04-11
Changed BME280 Module to option address

TomWS1:
Updated BME280 module so that the I2C address could be overridden with a connect() option.


gfwilliams:
Brilliant - thanks!

PR #502, opened 2019-04-14
Added basic support for the MAX3421E USB Host Shield

CWBudde:
I have created a basic header translation of Oleg Mazurov's MAX3421E headers. The original files can be found here: https://github.com/felis/ArduinoUSBhost/
The latest code from the original author offers by far more, but it's not that easy to translate as it uses templates and integrates more with access to several USB devices.
It would be necessary to write a similar library for Espruino, but that's by far more difficult.


gfwilliams:
Wow, thanks! This looks great - thanks for spending the time to do the documentation and image for it as well.

As mentioned in the forum thread (http://forum.espruino.com/conversations/332374/#comment14698027) I'll hold off merging for a few days as it feels like this is super close to being able to do something simple with a USB device.

PR #503, opened 2019-04-27
Plantower PMS7003 driver

AkosLukacs:
Driver for the Plantower PMS7003 PM sensor. It's a Chinese "dust" sensor. Actually a family of sensors, and according to info on the internet the code might work for others but I only have one type...
It's been working nicely on Espruino Wifi and ESP32 for some time.

One sidenote: honestly it's not entirely clear what's the difference between the "CF=1" and 'Atmospheric conditions" datasets returned by the sensor. Probably the "Atmospheric" variant should be used.

Oh, and a question: it's Chinese product so as far as I know there is no official english datasheet. Should I upload the datasheet I found somewhere, translated by someone?


gfwilliams:
Nice - thanks! And yes, if the datasheet isn't likely to stay in the same place then yes uploading would be great. If you stick in in the `datasheets` folder and link with eg. `[Datasheet](/datasheets/MAX7219.pdf)` then it should work fine.

AkosLukacs:
Added the datasheet + some comments in the code. I think it's ready to merge.

gfwilliams:
Nice - looks great, thanks! Mind if I change `PmsDriver.js` to `PMS7003.js` (or `PMSx003`)? It's how most of the other pages/modules are named so I think it'd make a lot of sense

AkosLukacs:
Ah, yes, I thought about renaming it too, but forgot it :|
Probably PMS7003 would be the better name, since afaik PMS3003 uses a different protocol. Already added PMS5003 to the docs, so if anyone is searching for it, they should find it.

gfwilliams:
Perfect - thanks! I'll get that done

PR #504, opened 2019-04-28
CCS811 improvements

AkosLukacs:
Hi, some improvements to the CCS811 module. Almost ready, but if you have some time, maybe take a look.
Done so far:
- support alternative I2C address
- support different sample rates
- use nWake to save power on lower sample rates
- moved the jsDoc from the constructor to the `.connect()` function, so the description will be visible on the website's reference section.
- and some docs
- of course my auto-format doesn't agree your auto-format. Do you have an editorconfig or something maybe?

Wip: Environmental data (temperature and humidity) compensation.

Btw, do you use this sensor?

### DIFF
--- a/devices/CCS811.js
+++ b/devices/CCS811.js
@@ -1,64 +1,94 @@
 /* Copyright (c) 2018 Gordon Williams, Pur3 Ltd. See the file LICENSE for copying permission. */
 
 var C = {
-  WHO_AM_I : 0x20,
-  WHO_AM_I_VALUE : 0x81,
-  
-  STATUS : 0x00,
-  STATUS_ERROR : 0x01,
-  STATUS_DATA_READY : 0x08,
-  STATUS_APP_VALID : 0x10,
-  STATUS_FW_MODE : 0x80,
-  
-  MEAS_MODE : 0x01,
-  MEAS_MODE_INT_THRESH : 0x04,
-  MEAS_MODE_INT_DATARDY : 0x08,
-  MEAS_MODE_DRIVE_MODE_IDLE : 0x00,
-	MEAS_MODE_DRIVE_MODE_1SEC : 0x10,
-	MEAS_MODE_DRIVE_MODE_10SEC : 0x20,
-	MEAS_MODE_DRIVE_MODE_60SEC : 0x30,
-	MEAS_MODE_DRIVE_MODE_250MS : 0x40,
-  
-  ALG_RESULT_DATA : 0x02,
-  
-  SW_RESET : 0xFF,
-  BOOTLOADER_APP_START : 0xF4,   
+  WHO_AM_I: 0x20,
+  WHO_AM_I_VALUE: 0x81,
+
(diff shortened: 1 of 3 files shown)


gfwilliams:
Looks good, thanks! - let me know when you're ready to merge.

I don't have any specific auto format style, no. It's basically K&R style with 2 spaces indent and curly braces on the same line, although what goes on inside the line itself isn't really prescribed. This might be interesting: http://www.espruino.com/FAQ#is-there-a-coding-style-i-should-use-

And yes, the sensor's in the Nordic Thingy:52 so this module gets included in the Espruino firmware for that - but it does mean that the module needs to stay backwards compatible :)

AkosLukacs:
Oh, there are tons of "small" things in Espruino you wouldn't even think about. Like what lines should be copy-pasted <3
It's still the same basic API, just "better" :)

AkosLukacs:
Added `setEnvData(humidity, temperature)` to use temperature and humidity compensation
Added `setMode(mode)` to change the sampling rate
Added firmware updater, so now there is a "native" Espruino solution to update the firmware in the CCS811. My main concern was that the 5k binaries wouldn't fit, but looks like it just works fine!

Messed with constants, but looks like closure compile doesn't want to inline them. (no-inlining in the original CCS811 code either)

Looks like there was a "something" in the [original code](https://github.com/espruino/EspruinoDocs/blob/master/devices/CCS811.js#L47): you read from register 0x80, but there is no 0x80 register in the datasheet. Or maybe there was, but didn't find it in either `[v1-00] 2016-Dec-23` or `[v1-06] 2019-Feb-07` (couldn't find other versions)

Ok, there is a slight breaking change (don't touch working code...): With fw 2.xx you can read the `isNew` information in one I2C read, but that doesn't work with fw 1.1.
Mainly that's why I created the firmware updater module. It's not small - 20k unminified, 1100-1200 vars usage in Pixl.JS. But that should still fit in all official boards, and even the ESP8266. The updater contains the v.2.0.0 and v.2.0.1 versions, and lots of info messages, so everybody should be able to upgrade in a couple of minutes.

AkosLukacs:
It's ready.
Tried to get closure compiler to inline the constants, but no luck. The module grew a bit, so manually shortened the constant property names, that saved 22 vars (roughly 10%).

gfwilliams:
That's great - thanks!

Sorry I missed your last comments for some reason. 

Yeah, splitting the object up into constants/regs as you've done is definitely better and usually means closure will inline at least some of it.

I need to look into why it's not touching constants though as it used to - I should maybe move to including a local closure compiler in the docs repo. It'd ensure everything stayed working at least and I guess might provide some info on why things weren't getting folded down

AkosLukacs:
No worries, at least had time for some more tweaks :)
Have you seen https://github.com/espruino/EspruinoDocs/pull/506?

gfwilliams:
I did, thanks! I'm a bit unsure at the moment but I'll post more info on there

PR #505, opened 2019-05-10
Added VL53L1X module

jahurtado:
This is a module for the laser ranging sensor `VL53L1X`. It is an evolution of the `VL53L0X` which can reach an accurate ranging up to 4 m.

The module has been adapted from https://github.com/pololu/vl53l1x-arduino/ trying to keep the compatibility with the `VL53L0X` module.


gfwilliams:
This looks great - thanks!

PR #506, opened 2019-05-14
Fixes for possible clearXXX bugs

AkosLukacs:
Wrapped most clearIntervals, clearWatches, clearTimeouts in if checks.
Some were already safe, but changed them for consistency.
Did not touch the `examples` folder, because the examples contained global `clearXXX()` calls, and didn't want to mess up.

The diff doesn't look so nice in GitHub due to whitespace changes. Set up my editor to remove line-end whitespaces, and hide in the diff. Sorry about that...

For reference: http://forum.espruino.com/conversations/333862

### DIFF
--- a/devices/DHT11.js
+++ b/devices/DHT11.js
@@ -26,8 +26,7 @@ DHT11.prototype.read = function (cb, n) {
   setTimeout(function() {pinMode(ht.pin,'input_pullup');pinMode(ht.pin);},20);
   // stop looking after 50ms
   setTimeout(function() {
-    clearWatch(ht.watch);
-    delete ht.watch;
+    if(ht.watch){ ht.watch = clearWatch(ht.watch); }
     var cks =
         parseInt(d.substr(2,8),2)+
         parseInt(d.substr(10,8),2)+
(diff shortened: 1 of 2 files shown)


gfwilliams:
Hi! Thanks for this - sorry for the lack of reply - I've been trying to figure out what I think about it.

I think with the recent changes to add exceptions for bad uses of `clearXX` hopefully any problems with the current code should come to light quite soon.

I know these changes are pretty straightforward, but I'm extremely wary of fixing stuff that isn't broken, and I imagine that very few of these changed modules have been tested?

And I know the changes are super minor and shouldn't really need testing, but literally every time I think that about something, turns out I broke it :)

AkosLukacs:
You are right about not touching not broken code!
Aaaand yes, only confirmed the error and and tested the fix with the DHT22. The DHT11 implementation is pretty much the same. I think this is as far as I can go with actually testing the changes.
Sorry for going overboard in the first place!

gfwilliams:
Thanks for all the work you put into this! Just merging your DHT changes

PR #508, opened 2019-05-24
Added eddystone UID module

renaudfv:
The **ble_eddystone.js** module is only valid for the URL version of the protocol. This PR adds **ble_eddystone_uid.js** which follows the same syntax, but allows sending _namespace_ and _instance_ byte arrays, providing also configurable interval setting.

It can be kept as a single file or could be included to the previous module. A naming syntax should be used though to make sure it does note break the functionnality for anyone.


gfwilliams:
Great, thanks!

PR #511, opened 2019-07-03
Docs - Update CCS811.md

AkosLukacs:
Wiring up table slightly broken on the last two cells:
image

JSDOC reference for the Firmware updater displays the main module's reference.


gfwilliams:
Thanks!

PR #514, opened 2019-07-20
Typo in graphical_menu.md?

AkosLukacs:
There is no `BTNU` / `BTND` / `BTNA` on the Pixl.
Of course it's not just for the Pixl, but now it "just works" on the Pixl.JS :)


gfwilliams:
Thanks - yes, that's because this was originally done for the Nodeconf 2017 badges and they had buttons arranged up/down/left/right/etc.

However iirc even those had the buttons numbered as well, so your change is definitely worth it :)

PR #515, opened 2019-08-01
var ll = require("NRF52LL"); missing from some snippets

AkosLukacs:
So copy-paste works for the lazy ones like me :)


gfwilliams:
Great - thanks!

PR #516, opened 2019-08-03
add module for the max44009 ambient light sensor

hinzundcode:


gfwilliams:
This looks great - thank you!

PR #517, opened 2019-08-04
NRF52 SAR ADC negative pin error

AkosLukacs:
A typo that breaks differential measurements.
That's why I prefer strongly typed languages over stringly typed ones, but still <3 Espruino :)


gfwilliams:
Brilliant - thanks for this - and for all the other little fixes/doc tweaks you've contributed recently!

PR #519, opened 2019-08-05
Added gain:1/4 and refvdd:true to the new sample

AkosLukacs:
Because otherwise it's pretty easy to jump out of the valid range (0.6V reference by default). And to match the other samples.

And the misleading about the over-range was that all values (at least for me) were 16380. And that's not the maximum valid value for 14 bits, but 4 lower. It was obviously wrong, but spent a couple of minutes trying to figure out what might be wrong...

On the positive side: It's working now :)


gfwilliams:
Thanks! For some reason I assumed `refvdd:true` was the default - it's not :)

PR #520, opened 2019-08-06
INA226 small fixes

AkosLukacs:
Fixes:
 - wiring table broken
 - vshunt calculation off by 10

Some tweaks:
 - `if (options.average) {` not needed, since always has value (the line above)
 - ~~`this.i2c.writeTo(this.addr,[a,d>>8,d&0xff]);` can be simplified to `this.i2c.writeTo(this.addr,[a,d]);`~~

### DIFF
--- a/devices/INA226.js
+++ b/devices/INA226.js
@@ -16,11 +16,9 @@ function INA226(i2c, options) {
   // config reg
   var config = 0x4127; // power on defaults
   options.average = options.average||256;
-  if (options.average) {
-    var i = [1,4,16,64,128,256,512,1024].indexOf(options.average);
-    if (i<0) throw new Error("average must be 1,4,16,64,128,256,512 or 1024");
-    config |= i<<9;
-  }
+  var i = [1,4,16,64,128,256,512,1024].indexOf(options.average);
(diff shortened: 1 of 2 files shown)


AkosLukacs:
I was wrong about the writeTo simplification, worked on the first try because I didn't power cycle the INA...

gfwilliams:
Yeah, you mean removing the 16 bit write? If you want to tidy it, `[a,d>>8,d]` works fine because the truncation to 8 bits is implicit

Are you sure about the `stop: false` as well? Most I2C devices are totally fine with stop being sent (and maybe expect it).  'if it ain't broke don't fix it' might apply to that...

AkosLukacs:
Oooops, I was testing if that would improve read speed, but didn't help, so
that's totally accidental. Don't push if you are in a hurry.
I will clean it up...

Gordon Williams <notifications@github.com> ezt írta (időpont: 2019. aug.
7., Sze 12:22):

gfwilliams:
Thanks!

AkosLukacs:
Ok, it's fixed now :)

gfwilliams:
Thanks!

PR #523, opened 2019-10-19
nfctag: fixing iOS 13 incompatibility

drandreas:
With iOS 13 Apple added support for writing NFC Tags.
However, the Apps I tried so far require the TAG to
be formated for NDEF. Unformatted tags are rejected.

The smallest valid NDEF message has an empty record
and an end marker 0xFE.


gfwilliams:
Brilliant - thanks!

PR #525, opened 2019-11-04
fix Uncaught SyntaxError

MaBecker:
Unexpected end of Input  at line 1 col 22
at line 1 col 22
Badge.getName = () =>


gfwilliams:
Did you actually get this error when uploading, or did you copy/paste into the left-hand side of the IDE?

Merging anyway though because it's just better your way :)

MaBecker:
Error occurred during upload on Pixl and B.....

Issue #526, opened 2019-11-21
Is there a reason 'nodejs' and not 'node' is used in the build?


mcdonnelldean:
This tripped me up trying to build the docs. I was going to submit a PR to change all references of `nodejs` to `node` which is the default. Before I do, is there a reason why it is no-standard?

gfwilliams:
Yes - it's because of this: https://launchpad.net/ubuntu/trusty/+package/nodejs-legacy

On some patforms (at least some linux, inc Raspberry Pi) `nodejs` is the default and you require a `nodejs-legacy` package (if it exists) to get `node`.

I guess really `nodejs` should be used everywhere if that's the one that is supported on all platforms - either way consistency would be good

mcdonnelldean:
Ok that makes sense! My only concern is if you are a JS dev and coming at this for the first time (eg you want to write a tutorial) you will get tripped up as most folks will have `node` not `nodejs` in their env.

However your point makes sense so it seems like an easy quick win here would be a change to the docs to export nodejs if you get an error. I will make a PR to the README for clarity and see what you think

gfwilliams:
Is `nodejs` actually not available on Mac/Windows by default? If so that probably outweighs the Raspberry Pi/Linux and we should just have a note to install `nodejs-legacy` to fix it.

Maybe someone with more knowledge can jump in here but maybe it's not an issue on newer distros - it's fine on Linux Mint 19.1 (based on Ubuntu 18.04) doesn't have a problem.

Personally I prefer `node` anyway.

mcdonnelldean:
@gfwilliams Yes indeed `node` is the default on Windows and Mac.

I've asked the question of a few friends who are using Linux to see what there thoughts are.

Happy to make a PR to change it to `node` but it would be nice if it self shimmed, let me come back to this later this evening to see if I can get a nice cross compat shim in so it becomes moot.

I suspect falling down here at first install has likely put a lot of wouldbe contributors off so i'll PR something this evening, even if it's just a notice.

mcdonnelldean:
@gfwilliams Thinking about this a little more, given it's a mess to change alias' for the average Windows user but not for a Linux user, I think anote as you mentioned may be the way to go, let me craft a PR and see what your thoughts are

gfwilliams:
Thanks - sounds good to me.

The hope is that you can contribute to the documentation without having to get the whole toolchain working though, using just the GitHub Markdown preview.

[...]

PR #528, opened 2019-11-22
Update install experience

mcdonnelldean:
Based on the discussion on #526:

I've opted to default to node to serve the majority on the basis that a user that knows Linux is far more likely to be able to run a symlink command than a user on say, Windows.

I've made other minor improvements is well, all in the vein that if someone does pull the code to build and view locally, it will run as expected with minimal effort or hiccup.

Please note the two files I ignored. I didn't see a package-lock checked in so I opted to ignore it, same for ordering.txt which is generated post build run.

Notes are:

- Change nodejs to node in all places
- Update docs to provide info on how to symlink to nodejs
- add some build bits to gitignore to prevent bad check-ins
- add redirect in start script to handle requests to /

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ references.json
 /.project
 npm-debug.log
 orig
+ordering.txt
+package-lock.json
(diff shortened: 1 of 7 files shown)


gfwilliams:
Looks great - thanks!

PR #531, opened 2019-11-28
Emit error on first time MQTT connection failure

mcdonnelldean:
Emit an error if the client did not connect the first time. For instance, if the IP address is wrong you will get `Unable to create socket`. This is helpful so that disconnect / reconnect logic can take into account bad server ip without guessing the problem.


gfwilliams:
Neat - thanks!

PR #534, opened 2020-01-12
Converter: heatshrink_compress should be heatshrink.compress

AkosLukacs:
Doesn't work, I guess a typo...


MaBecker:
Wow, so you immediately tested it :)

gfwilliams:
Great - thanks!

PR #540, opened 2020-02-25
SX127x - RFM95W and TTGO LoRa32 transmission fix

AkosLukacs:
This is kind of complicated: there is a low-power and a high power amplifier on the SX127x chips, but apparently Hope RF and TTGO didn't connect the low-power amplifier's output RFO pins to the antenna output. So if anybody wants to transmit must use the PA_BOOST output to get any RF signal.
Of course this is not documented: Hope RF's datasheet is just a copy-paste of the Semtech datasheet, and there is no datasheet at all for the TTGO module.

The only clue I could find is a [blog post](https://www.disk91.com/2019/technology/lora/hoperf-rfm95-and-arduino-a-low-cost-lorawan-solution/) and comments in RadioHead library's RH_RF95.cpp like this:

        // RFM95/96/97/98 does not have RFO pins connected to anything. Only PA_BOOST
        // pin is connected, so must use PA_BOOST

AFAIK you can't detect what module is connected, to to work around this, added a new optional config parameter: `forcePaBoost`.
Tried to follow the origins of the power register calculation back to the Semtech sample, but looks like that code depended on their test board's pin being pulled to 1 or 0 to determine which board they have. So did the power, power DAC and current limit code from scratch.

On the boards I have tested a couple of TX power settings, and the RSSI did change, so assume it's working.

Caveat: I can't test with "proper" SX127x module like the Modtronix one, because I only have the TTGO and HopeRF modules.

Caveat 2: Checked three libraries, and they all have different lower power limits and different register value calculation on the lower end. No idea if there is a reason for the different limits, or just to keep their code simpler?
 - Semtech says -4dBm is the lowest output power. But if my math works, you can't configure exactly -4dBm. You can set it to -4.2, -3.6, -3.0dBm???
 - [Sandeep Mistry's Arduino lib](https://github.com/sandeepmistry/arduino-LoRa) limits power to 0dBm.
 - [RadioHead 1.98](http://www.airspayce.com/mikem/arduino/RadioHead/) limits power to -1dBm
 - [RadioLib](https://github.com/jgromes/RadioLib/blob/master/src/modules/SX127x/SX1278.cpp) limits to -3dBm. Actually this is the lowest non-fractional value you can set.

Can't test these settings with the HopeRF and TTGO modules because +2dBm is the lower limit using PA_BOOST...

### DIFF
--- a/devices/SX127x.js
+++ b/devices/SX127x.js
@@ -577,31 +577,36 @@ SX.prototype.setRxConfig = function( options ) {
 
 SX.prototype.setTxConfig = function( options ) {
   options = this.commonSetConfig(options);
-  var paConfig = this.r( REG.PACONFIG );
-  var paDac = this.r( REG.PADAC );    
-
-  paConfig = ( paConfig & RF.PACONFIG_PASELECT_MASK ) | RF.PACONFIG_PASELECT_RFO;
-  // TODO: if 525000000 and  SX1276MB1LAS , use PABOOST, not RFO
-  paConfig = ( paConfig & RF.PACONFIG_MAX_POWER_MASK ) | 0x70;
-
-  if (paConfig & RF.PACONFIG_PASELECT_PABOOST) {
-    if (options.power > 17) 
-      paDac = ( paDac & RF.PADAC_20DBM_MASK ) | RF.PADAC_20DBM_ON;
-    else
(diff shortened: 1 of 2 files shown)


gfwilliams:
This looks great - thanks! Especially for the added/improved docs!

PR #542, opened 2020-03-22
ST7789 LCD driver & documentation

AkosLukacs:
Normal mode works, tested on TTGO T-Display (ESP32 with built-in LCD) board.

https://github.com/espruino/EspruinoDocs/blob/4e0a6dc280cb360bd70e0326863623647f78ac62/devices/ST7789/ST7789_module.jpg

Most likely only 135x240 resolution works, as other resolutions seem to require different offsets, but don't have other LCDs to test.

Paletted mode (commented out for now) produces some interesting result: 
image


gfwilliams:
Looks great, thanks!

Paletted issues will likely be because there's some mismatch between the amount of pixels sent per row and the 'row width' sent - just an out by 1 error here maybe? https://github.com/espruino/EspruinoDocs/pull/542/commits/4e0a6dc280cb360bd70e0326863623647f78ac62#diff-eca351fa8c72bb37044e1ca2652c1ef6R121

I'll merge this now but if you get a chance to tweak it, it'd be really cool.

PR #546, opened 2020-03-28
Add clarification for beep mode

Xaseron:


gfwilliams:
Thanks - actually 'Vibrate' should be the default anyway now once settings is installed, but yes, I'd forgotten you need to add Settings too!

PR #551, opened 2020-04-14
added 'Kafka tiny client'

sqlly:
I've created a most lightweight implementation of Kafka binary protocol for Espruino. Kafka is now almost de-facto standard for stream messaging and Espruino definitely lack such implementation.

### DIFF
--- a/modules/kafka-tiny-client.js
+++ b/modules/kafka-tiny-client.js
@@ -1,0 +1,540 @@
+/* Copyright (c) 2019 Pavel Kutakov. See the file LICENSE for copying permission. 
+  
+Tiny Kafka Client
+
+
+This is a simplest implementation of Kafka client. 
+It is based on the lowest compatible version of Kafka binary protocol to reduce message size.
+Can be used only to publish messages to Kafka which is pretty typical scenario in IoT world.
+
+
+
+Short Example:
+---------------------------
+const Kafka  = require('kafka-tiny-client');
+ 
+const kafka = new Kafka({
+  client: 'my-app',  //Will take device serial number when not specified
+  host: '130.61.255.148',
+  port: 9092,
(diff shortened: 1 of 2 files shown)


gfwilliams:
This is great, thanks! Does the crc32 code you have there actually work?

On new Espruino builds (cutting edge, and 2v05 when released) there's actually a `E.CRC32` function which would probably help out: http://www.espruino.com/Reference#l_E_CRC32

If you need it I also have an implementation that doesn't need the table...

sqlly:
Probably my Espruino build was not so fresh (actually I've created that code 3 months ago) and at that time I've found that build-in CRC32 value does not equal to what Kafka expects. After a deep dive to that I've found working solution.
As for implementation that does not need a table - I've tried one with dynamic CRC table generation - it was extremly slow. If your version is really fast, I will be glad to remove CRC table, it is very resource expensive.

gfwilliams:
This should do it hopefully:

```
function CRC32(data) {
  var crc = 0xFFFFFFFF;
  data.forEach(function(d) {
    crc^=d;
    crc=(crc>>>1)^(0xEDB88320&-(crc&1));
    crc=(crc>>>1)^(0xEDB88320&-(crc&1));
    crc=(crc>>>1)^(0xEDB88320&-(crc&1));
    crc=(crc>>>1)^(0xEDB88320&-(crc&1));
    crc=(crc>>>1)^(0xEDB88320&-(crc&1));
    crc=(crc>>>1)^(0xEDB88320&-(crc&1));
    crc=(crc>>>1)^(0xEDB88320&-(crc&1));
    crc=(crc>>>1)^(0xEDB88320&-(crc&1));
  });
  return ~crc;
}
print(CRC32([2,4,65,7,5,4,3,5,6]))
```

PR #552, opened 2020-05-04
Add Teletext-style / BBC Micro MODE 7 fonts

tomgidden:
These fonts are a little on the large side, so I've used heatshrink for storing the double-height versions.  They do not include the upper-bit graphics characters.


gfwilliams:
That's awesome - thanks! Can't wait to see something using this :)

tomgidden:
Thanks @gfwilliams... I have a watch face in development that uses it. It's a little too inefficient to push yet though.

Is there a delay in updating www.espruino.com/modules ? I assumed it was an automatic deployment.

All the best,
Tom

gfwilliams:
No, I manually update it. Am doing it now but it's taking a while as I need to cope with heatshrink in fonts when the preview images are made ;)

tomgidden:
Heh.. oops! Sorry about that :)

Incidentally, can you think of a simpler way than
```
const font = new Uint8Array(require('heatshrink').decompress(atob("…"))).reduce((p,c)=>p+String.fromCharCode(c),"");
```
for passing to `setFontCustom`?  I tried some other more straightforward ways than the `reduce` string construction, but didn't have much luck.

gfwilliams:
Oh wow, didn't spot that. That's going to be crazy slow and possibly not great for memory usage.

`E.toString(require('heatshrink').decompress...)` is what you need :) I'll update them now

gfwilliams:
They are actually now live (old versions) if you want to try...

tomgidden:
Thanks...  I would, but my Bangle is stuck "Searching for GPS time". That's for another forum, though.

gfwilliams:
> "Searching for GPS time"

https://www.espruino.com/Troubleshooting+Bangle.js#my-bangle-is-stuck-showing-searching-for-gps-time-

Likely by the time you get it sorted, the `toString` version will be live :)

PR #554, opened 2020-05-13
update ws5100

MaBecker:
please double check - thanks


gfwilliams:
Ahh - looks like the issue is the directory is `devices/adruino-w5100` and not `devices/arduino-w5100` :)

MaBecker:
Thanks!

gfwilliams:
Thanks for fixing it :)

PR #559, opened 2020-06-01
Firmware Updates: Activate location in Android 10

ikkala:
I deploy new firmware so rarely to my Bangle.js that I always forget this little detail, so I though someone else could also use this info.

Aligned also the bullets, they seem to be twisted here: https://www.espruino.com/Bangle.js#firmware-updates

(sorry the bad commit message "Firmware Updates: Activate location activation")


gfwilliams:
That's great - thank you!

PR #569, opened 2020-08-17
update the motor shield documentation and code

Isanderthul:
I could place this in the forum, tutorials section if you feel that it would be better suited there (@gfwilliams)


gfwilliams:
Hi,

Thanks! Do you mind if I move the license text to just a single line? The LICENSE file in the repo is MIT so it's the same, it's just good to keep all the modules in basically the same form.

Also, are you sure that setting M1A and M1B causes a short? It seems to be it just makes both motor outputs positive, which is basically the same as connecting both to 0v - there's nothing too bad about that.

PR #571, opened 2020-08-21
add custom stuff

MaBecker:
custom board
custom lib spi lcd unbuf


MaBecker:
Hope I covered all things that are required for a tutorial about this.

PR #576, opened 2020-09-04
Fix bactick syntax

Wilfred:
This currently isn't displaying correctly: https://www.espruino.com/FAQ#when-i-type-a-command-why-does-espruino-print-undefined-


gfwilliams:
Ouch, thanks! Not sure how we missed that!

PR #579, opened 2020-09-29
Add whitespace to font

berkenbu:
Added whitespace (char code 32) to font, had to fix some small issues with the generator code.


gfwilliams:
That's brilliant - thanks!

PR #580, opened 2020-10-20
Create MLX90615.js

jolielerner99:
Using MLX90615 to read temperature through Puck.js for class.


gfwilliams:
Nice, thanks!

Please could you paste up a few lines of example code using this module here so I can add some simple docs for this?

PR #586, opened 2020-11-13
add Font4x5

MaBecker:
Simple 4x5 font including source in a separate file.


gfwilliams:
Thanks - I'll stick the source code in the comments at the head of the file (I've done this on some other fonts like the 7 segment ones).

Also just tested and you can't just supply `' '`as the start char code - it should be a number, so I've fixed that for the EspruinoDocs version

MaBecker:
Thanks

MaBecker:
thanks for  updating Fonts.md too.

PR #587, opened 2020-12-08
Add DANE Arwes module

OmegaRogue:
Added a module for UI based on https://arwes.dev for use in my BangleJs Apps

### DIFF
--- a/modules/dane_arwes.js
+++ b/modules/dane_arwes.js
@@ -13,6 +13,7 @@ function Arwes(cornerSize, cornerOffset, borderWidth) {
   this.cornerSize = cornerSize;
   this.cornerOffset = cornerOffset;
   this.borderWidth = borderWidth;
+  this.color = this.C.color;
 }
 
 
--- a/modules/dane_arwes.js
+++ b/modules/dane_arwes.js
@@ -110,7 +110,7 @@ Arwes.prototype.drawFrame = function (x1, y1, x2, y2) {
   drawTopRightCorner(this, x2, y1);
   drawBottomLeftCorner(this, x1, y2);
   drawBottomRightCorner(this, x2, y2);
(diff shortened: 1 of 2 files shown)


gfwilliams:
Hi - this looks good - however, is the intention to add more to this in the future?

It seems overkill to add a module to Espruino that basically just draws 9 rectangles - I'm not sure it's going to get a bunch of use outside of your specific apps. Maybe I could find a way to define modules in BangleApps that can be used there?

OmegaRogue:
> Hi - this looks good - however, is the intention to add more to this in the future?
> 
> It seems overkill to add a module to Espruino that basically just draws 9 rectangles - I'm not sure it's going to get a bunch of use outside of your specific apps. Maybe I could find a way to define modules in BangleApps that can be used there?

I'm currently working on adding versions of functions like E.showAlert, E.showMenu, E.showMessage and E.showPrompt to it,
but I'm not expecting anyone else to really use it, so being able to define modules in BangleApps would probably be the better option

OmegaRogue:
I've tried to compile all the functions in the module, but the IDE said `WARNING: Function marked with "compiled" uploaded in source form` and then couldn't find the functions when they were used.

gfwilliams:
> I've tried to compile all the functions in the module

Was that in the emulator or a Bangle? 

Honestly trying to compile the functions wouldn't have been a great idea anyway though. There are a few of things (I mentioned some in your original post) that should speed the code up.

OmegaRogue:
> > I've tried to compile all the functions in the module
> 
> Was that in the emulator or a Bangle?
> 
> Honestly trying to compile the functions wouldn't have been a great idea anyway though. There are a few of things (I mentioned some in your original post) that should speed the code up.

it was on both a bangle and on the emulator

gfwilliams:
Thanks - I'll see about tweaking the app loader to let it pull in modules now.

PR #588, opened 2020-12-24
Update tinyMQTT.js

MaBecker:
- call onDat if data has more than one payload
- return only message


gfwilliams:
Nice - thanks!

There was another issue which is what happens when an MQTT command is split over multiple 'data' events - but this will at least solve most problems for  ESPxx users.

PR #589, opened 2020-12-31
Remove NDEFWriter

beaufortfrancois:
This PR updates Web NFC API to reflect latest spec changes.
See https://github.com/w3c/web-nfc/blob/gh-pages/EXPLAINER.md#changes-done-after-origin-trial-ot and https://web.dev/nfc/

Note that the first call `ndef.scan()` will fail if permission was not granted before as a user gesture is required.

@gfwilliams

### DIFF
--- a/tutorials/Web NFC.md
+++ b/tutorials/Web NFC.md
@@ -93,12 +93,11 @@ Then you can click the 'try now' link at the bottom of the page below to try it 
 <script>
 /* Starting a scan at startup stops Android from
 moving away from the Chrome window when a tag is found*/
-const reader = new NDEFReader();
-const writer = new NDEFWriter();
-reader.scan();
+const ndef = new NDEFReader();
+ndef.scan();
 
 function send(msg) {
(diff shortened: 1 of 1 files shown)


gfwilliams:
Great - thanks!

So I guess right now you and up having to do something like:

* Load page
* Click 'write'
* Accept permission prompt
* Reload page
* Everything works fine?

beaufortfrancois:
> So I guess right now you and up having to do something like:
> 
> * Load page
> * Click 'write'
> * Accept permission prompt
> * Reload page
> * Everything works fine?

Yes! Ideally, I'd go with something like https://web.dev/nfc/#check-for-permission

beaufortfrancois:
For info, https://www.espruino.com/Web+NFC#web-page is still not updated. Is this "normal"? @gfwilliams

gfwilliams:
Thanks!

And yes - I manually pull in the changes from GitHub, then review and update the site. It helps to stop things getting broken ;)

PR #590, opened 2021-01-03
Improvement: Update minify.js

MaBecker:
definittly improves minification - thanks to allObjects

### DIFF
--- a/bin/minify.js
+++ b/bin/minify.js
@@ -55,7 +55,8 @@ function closureOnline() {
     ['output_info','warnings'],
     ['warning_level','VERBOSE'],
     ['js_code',js],
-    ['js_externs',jsExterns]
+    ['js_externs',jsExterns],
+    ['language_out','ECMASCRIPT_2015'],
    //['language','ECMASCRIPT5_STRICT'],
   ];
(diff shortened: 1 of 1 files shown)


gfwilliams:
By improves, you mean it produces more compact code, or it fixes issues where bad code was created?

MaBecker:
was impresssed by the more compact code.

MaBecker:
was using this page for tests

https://closure-compiler.appspot.com/home

MaBecker:
Just did some more research, there is to much risk to merge this as there are many min.js files that are generated by minify.js.
So better close it an try to find a way to pass additional compiler flag by command line or conf file.

gfwilliams:
Well, this might still be ok - really it's just whether it tries to use any features Espruino doesn't support.

Sorry for the slow replies but I'm working through hundreds of emails after Christmas!

I believe it'd be pretty easy to add it to what's there already. Right now we already look to see if there's a comment inside the file that mentions only whitespace optimisation should be applied: https://github.com/espruino/EspruinoDocs/blob/master/bin/minify.js#L44

MaBecker:
Ok, so let's open this again and add something.

PR #592, opened 2021-01-08
Update buttons visibility when scan starts

beaufortfrancois:
buttons visibility should be updated only when scan starts successfully.

@gfwilliams


gfwilliams:
Thanks!

PR #593, opened 2021-02-04
Added diagram illustrating pin-out for charging pins on Bangle.js

omgmog:


gfwilliams:
Great - thanks for this!

How did the DIY charger go?

omgmog:
> Great - thanks for this!
> 
> How did the DIY charger go?

Currently 3D-printing a jig to hold some cables in the right position, then I'll let you know!

PR #594, opened 2021-02-24
Add module StringDecoders

MaBecker:
Is can be a pool for special string decoding functions like fitWords().


gfwilliams:
Nice - thanks!

PR #597, opened 2021-03-28
Fix typo

Wilfred:


gfwilliams:
Thanks for this - these tweaks are really handy!

PR #599, opened 2021-04-23
Adds support for the SHT4x humidity and temperature sensor

gniezen:
I based this heavily off of the [SHT3C module](https://github.com/espruino/EspruinoDocs/blob/master/devices/SHT3C.js), but with the necessary changes as described in [the SHT4x datasheet](https://cdn-learn.adafruit.com/assets/assets/000/099/223/original/Sensirion_Humidity_Sensors_SHT4x_Datasheet.pdf?1612388531).


gfwilliams:
Thanks! That's great! I'll merge now, but it might be a few days before I update the website and bring this in

PR #601, opened 2021-06-13
Update DS3231.md

sleuthware:
Add link to module source: /modules/DS3231.js


gfwilliams:
Thanks - I just committed a fix for this rather than pulling in directly.

Ideally you should put markdown in the `.md` file (eg https://www.espruino.com/Writing+Modules) as opposed to HTML

PR #616, opened 2021-11-01
graphical_menu improvements

rigrig:
- remove unused parameter from `select()`
- shave a few bytes off `move()`

If a numerical item has `item.wrap` we no longer check if `item.min/max` is set, because your menu will be broken either way.


gfwilliams:
Thanks!

PR #618, opened 2021-11-28
Update README to include EspruinoWebTools

dandelany:
I had to check out both `Espruino` and `EspruinoWebTools` repos in order to get build.sh to run.


gfwilliams:
Thanks!

PR #621, opened 2021-12-01
Improve font converter - allow font file selection

dandelany:
I wanted to use the [Espruino Font Converter](http://www.espruino.com/Font+Converter) with a local .otf font file I had - and I realized it was possible to do it in Javascript by generating a Base64 data URL with `FileReader`. Hope this is useful to others too :) Feel free to make any changes.

PS: FYI I have noticed a separate bug - when using Google fonts, the font converter sometimes creates the font *before* the font loads, and you have to click "Go" again to regenerate it - but this only happens occasionally & I haven't tracked it down yet.

### DIFF
--- a/info/Font Converter.md
+++ b/info/Font Converter.md
@@ -12,31 +12,63 @@ with Espruino's [Graphics](/Graphics) library.
 
 How it works:
 
-* Go to [https://fonts.google.com/](https://fonts.google.com/), find a font
-* Click `Select This Style`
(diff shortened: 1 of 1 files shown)


gfwilliams:
That's brilliant - thanks! Not sure if you spotted this but someone was asking just today about it! http://forum.espruino.com/conversations/370540/#comment16291655

I'll try and update the website later today with this

PR #622, opened 2021-12-01
Bangle.js settings tutorial: remove references to Settings library

rigrig:
My [first version](https://github.com/espruino/EspruinoDocs/blob/1820bb2822deefcb22dff2bd06edabfff5ceffd1/tutorials/Bangle.js%20App%20Settings.md) had a bunch more snippets with different ways to read settings, I settled on this as the neatest way to specify a list of defaults:

```JS
var settings = Object.assign({
  // default values
  something: 123,
  anotherthing: 456,
}, require('Storage').readJSON("myapp.json", true) || {});
```

Also added some comments to the example about the `!!` and `0|` trickery, and fixed the example "data" filename.


gfwilliams:
Thanks! Object.assign is *way* cleaner

PR #623, opened 2021-12-08
Update graphical_menu.md

myxor:
Update example of `menuinfo` object so that  it's syntax is valid and therefore it can be used in emulator


gfwilliams:
Neat - thanks!

PR #625, opened 2022-01-07
Add bluetooth combination HID interface.

7kasper:
As mentioned in [https://github.com/espruino/BangleApps/pull/1199](https://github.com/espruino/BangleApps/pull/1199).

Adds a BLE HID interface that supports:

- [x] Working mouse on windows 10 (default mouse doesn't seem to work)
- [x] Mouse with 5 buttons
- [x] Verticle and horizontal scroll
- [x] Dragging action
- [x] Keyboard at the same time
- [x] Pressing up to 5 concurrent keys
- [x] Holding and releasing of keys.

I haven't tested the concurrent holding of keys yet nor the complete library but I think its correct this way.. 😸 
The HID report itself has been tested as it was developed and used for the presentor app [tested on banglejs2, but working shouldn't differ from banglejs1 or puckjs I think].

Feel free to comment or suggest changes.
EG: I was uncertain whether or not to use promises instead of callbacks so you could more easily chain actions.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = yes
+
+[*]
+indent_size = 2
+indent_style = space
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,0 @@
-root = yes
-
-[*]
-indent_size = 2
-indent_style = space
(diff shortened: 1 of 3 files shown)


gfwilliams:
Nice - this looks great - thanks!

I notice you've got a `.editorconfig` file committed though? Please could you delete that and I can get this merged

7kasper:
Sorry! Its gone now.

gfwilliams:
Thanks!

PR #626, opened 2022-01-11
Bangle.getOptions ref to setOptions

DDDanny:
Bangle.getOptions currently refer to itsself
I'll continously try to correct the docs when I discover typos etc.

### DIFF
--- a/bin/espruino.json
+++ b/bin/espruino.json
@@ -2859,8 +2859,8 @@
     },
     "getOptions": {
       "!type": "fn() -> ?",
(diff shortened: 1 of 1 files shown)


gfwilliams:
Hi - thanks! This is actually auto-generated from the code. I've just fixed this (see the referenced commit above) so I'll close this and the JSON (and reference) will update automatically next time I push changes

DDDanny:
Allright 👍 I see - like "javadoc" or "swagger" etc. the documentation is generated from source code annotations - I'll then pull request the correct repro / file

PR #629, opened 2022-01-28
Update Bangle.js docs because 'About' is now called 'More...'

jes:
For https://github.com/espruino/BangleApps/pull/1364#issuecomment-1024204425


gfwilliams:
Thanks! You're fast! I was on this already :)

PR #631, opened 2022-02-03
GadgetBridge: battery "status" message now also includes `chg`

rigrig:


gfwilliams:
Thanks for updating the docs!

PR #632, opened 2022-02-17
Update Bangle.js 2 guidance for iOS users

nh-99:
There's currently a bug affecting WebBLE that makes it so the Bangle.js 2 doesn't show up in the list of devices to connect to if it's already connected. This adds a note about it to the documentation.


gfwilliams:
Thanks! It'll update next time I push changes to the website

PR #635, opened 2022-03-04
Use sleep/wakeup commands for PMS7003 sensor

gniezen:
Newer releases of the PMS7003 sensor don't power down when the RESET or SET pins are pulled low, so we have to use the sleep/wakeup commands if we want to save energy.

I've removed the references to the RESET and SET pins are they're not relevant anymore, and replaced it with the sleep/wakeup commands.


gfwilliams:
That looks great- thank you!

PR #640, opened 2022-04-13
Font Converter: Attempt to add heatshrink compression

KTibow:
This attempts to add compression with Heatshrink as an option to the font converter. Check the diff.
Tested for me and works. A broken preview is [available](https://ktibow.github.io/EspruinoDocs/info/Font%20Converter) as long as you paste in the contents of https://www.espruino.com/js/heatshrink.js into the console first.


gfwilliams:
This looks great - thanks!

PR #646, opened 2022-05-26
DRV2605 haptic motor controller

ClintH:
Port of Ladyada's Arduino library, with a few extras.


gfwilliams:
Looks great - thanks!

PR #656, opened 2022-07-04
Clarifying notes on debugging the Bangle.js 2 with the nRF52 DK

freetink3r:
The nRF52DK.png image was incorrectly named, so the image isn't showing up on the page. The image was also missing the connection between the nRF52 DK `VTG` and `VDD` pins, which are required to be shorted on the latest boards in order to debug an external board like the Bangle.js 2. This commit removes the old nRF52 image (which wasn't showing up anyway because of the file extension mismatch) and adds a new schematic showing the additional short required.


freetink3r:
UPDATE: The silkscreen on my board does match the old image, with `VDD` next to `VTG`, but Nordic's documentation shows something different in their documentation: 
image

I pulled that from here: https://infocenter.nordicsemi.com/topic/ug_nrf52832_dk/UG/nrf52_DK/hw_drawings_v1.0.html

At any rate, the external board definitely doesn't work for me without connecting `VTG` to `VDD` or `VDD nRF` (either of the two pins next to `VTG`).

PR #657, opened 2022-07-08
Fixes some typos

fewieden:


gfwilliams:
Thanks! Stuff like this is a great help as it never gets reported otherwise :)

PR #658, opened 2022-07-09
Allow delay arg for type() function

BraidenPsiuk:
Helpful for emulating keyboards on older, slower devices

### DIFF
--- a/modules/USBKeyboard.js
+++ b/modules/USBKeyboard.js
@@ -191,7 +191,7 @@ exports.tap = function(key, callback) {
   }, 10);
 }
 
-exports.type = function(txt, callback) {
+exports.type = function(txt, delay, callback) {
   var intr = setInterval(function() {
     if (!txt.length) {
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks, but I'm afraid this will break anyone's code that currently uses `USBKeyboard` because the arguments changed.

Do you think you could add delay as the last argument so it'll still work? Something like:

```
exports.type = function(txt, callback, delay) {
  delay = delay||20;
  var intr = setInterval(function() {
    if (!txt.length) {
      clearInterval(intr);
      if (callback) callback();
    } else {
      if (txt[0] in KEY) exports.tap(KEY[txt[0]]);
      txt = txt.substr(1);
    }
  }, delay);
}
```

BraidenPsiuk:
Yes, my bad, thanks for catching that!

gfwilliams:
Great - thanks!

PR #660, opened 2022-08-10
Bangle.js Gadgetbridge is on F-Droid

rigrig:
* `Bangle.js Gadgetbridge` is now available on F-Droid: add+update links, remove text about F-Droid version limitations
* The `Gadgetbridge` project is still mentioned, but all app links now point at the `Bangle.js Gadgetbridge` flavor, so we can be a little less verbose about functionality like `http` only existing for that flavor.


gfwilliams:
Argh, I literally just added some changes like this (which I didn't commit yet) :)

gfwilliams:
closed with https://github.com/espruino/EspruinoDocs/commit/44f20731474274e693064ed087a25043e9203798

PR #662, opened 2022-08-17
Gadgetbridge.md updated re intents

thyttan:
... to describe extended functionality.

Should it be updated now or wait until the referenced functionality is present in a published version of Gadgetbridge?


gfwilliams:
Great, thanks! I've just pushed to the play store, so merging this now

PR #665, opened 2022-08-23
added support for ADS1115 in ADS1X15 module

yerpj:
Based on this discussion : http://forum.espruino.com/conversations/378555/#comment16639545


devices/ADS1X15.js
```diff
@@ -92,10 +93,12 @@ var REG = {
   LOWTHRESH : 2,
   HITHRESH : 3
 };
-function ADS1X15(i2c) {
+function ADS1X15(i2c,opts={part:"ads1015"}) {
```
gfwilliams, line 96: I'm afraid Espruino doesn't support default arguments, so I don't believe this code will even run on Espruino unless the minifier is smart enough to transpile it

PR #672, opened 2022-10-10
Create fetch.js

JThobbyist:
Minified Fetch.js File.

Usage: 

`const fetch = require('fetch.js')(http_module?)`
`fetch('http://google.com').then(r=>r.text()).then(console.log)`

`http_module` argument is optional, to specify which module to use under-the-hood currently only the 'http' module is supported, and if the argument is omitted it will default to 'http' anyways.

In the future I hope to add support for the 'tls' module.


gfwilliams:
Thanks! I'll merge but will swap this over to the unminified version - when we publish we automatically produce a minified version from unminified source

gfwilliams:
Just noticed:

> http_module argument is optional, to specify which module to use under-the-hood currently only the 'http' module is supported, and if the argument is omitted it will default to 'http' anyways.

Is there any other module in Espruino that could be used? As far as I know, `http` will automatically switch to `https` if it's in the URL and the build supports it... If Espruino can't support anything other than `http` then maybe we should comment that bit out for now so it's not wasting code

JThobbyist:
I thought only the ‘tls‘ module supported HTTPS (and that it needed the certs in the argument?). If the ‘http’ module also can do HTTPS, then there’s no need to use any other module. We can even just export the fetch function itself using the ‘http’ module instead of the argument from the wrapper. That’s actually how I originally structured it, but I thought it wouldn’t be able to do HTTPS. -Joe TeglasiOn Oct 12, 2022, at 3:41 AM, Gordon Williams ***@***.***> wrote:﻿
Just noticed:

http_module argument is optional, to specify which module to use under-the-hood currently only the 'http' module is supported, and if the argument is omitted it will default to 'http' anyways.

Is there any other module in Espruino that could be used? As far as I know, http will automatically switch to https if it's in the URL and the build supports it... If Espruino can't support anything other than http then maybe we should comment that bit out for now so it's not wasting code

gfwilliams:
Ok, just confirmed: http get calls jswrap_net_connect (https://github.com/espruino/Espruino/blob/1e7007d15/libs/network/http/jswrap_http.c#L430) and that does a check for `https`: https://github.com/espruino/Espruino/blob/1e7007d15/libs/network/jswrap_net.c#L402

If you want to specify certs then you just supply them as options to `http.request`: http://www.espruino.com/Storing+HTTPS+Certificates

I'd take the extra code out then - because `tls` is for a standard socket connection (eg no HTTP handling/parsing) so couldn't be used anyway

PR #673, opened 2022-10-10
Create fetch.md

JThobbyist:
goes with fetch.js

### DIFF
--- a/modules/fetch.md
+++ b/modules/fetch.md
@@ -1,0 +1,126 @@
+# Simple Fetch-like API for Espruino
+created by Joe Teglasi, Copyright 2022, License: MIT
+
+
+#### Initiation:
+```
+const http = require('http');
+const fetch = require('fetch.js')(http);
+```
+OR
+```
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks!

gfwilliams:
Ahh - didn't spot this. There's a specific format for docs - see the other ones in the repo, and: http://www.espruino.com/Writing+Modules

I'll modify it now too

JThobbyist:
Ah okay. I didn’t see that. I will use that format moving forward. Do you want me to fix the ones I uploaded and make a new PR?-Joe TeglasiOn Oct 12, 2022, at 3:42 AM, Gordon Williams ***@***.***> wrote:﻿
Ahh - didn't spot this. There's a specific format for docs - see the other ones in the repo, and: http://www.espruino.com/Writing+Modules
I'll modify it now too

gfwilliams:
Thanks! But no, that's fine, I fixed it this morning.

If you wanted to remove the optional http provider it might be worth doing a PR for that though :)

PR #677, opened 2022-10-12
Update fetch.js

JThobbyist:
fixed issue with request-promise not rejecting on aborted request (using AbortController.js)


gfwilliams:
Thanks!

PR #680, opened 2022-11-01
update shoplink

x29a:
so it works and also remove outdated price information

### DIFF
--- a/boards/Bangle.js2.md
+++ b/boards/Bangle.js2.md
@@ -11,7 +11,7 @@ Bangle.js is the successor to [Bangle.js](/Bangle.js), and is based on the
 
 [![](Banglejs2/main.jpg)](https://www.kickstarter.com/projects/gfw/banglejs-2-the-open-smart-watch)
 
-* BUYFROM: £64,£49,https://shop.espruino.com/banglejs2,/Order#banglejs2
+* BUYFROM: https://shop.espruino.com/banglejs2
 
 **Bangle.js 2 is an open, hackable smartwatch**
 


gfwilliams:
Hi - please can you explain why this doesn't work? I just tried https://www.espruino.com/Bangle.js2 and it's fine. The second line for distributors works too.

Pricing is also correct, it's just without VAT - which would have to be paid in UK/EU not not elsewhere

x29a:
@gfwilliams hi and thanks for checking in so quickly.

when i open the original link [https://shop.espruino.com/banglejs2,/Order#banglejs2](https://shop.espruino.com/banglejs2,/Order#banglejs2) i get a:

The page you requested cannot be found!

From the shop. Tried it with Chrome and Firefox on Ubuntu and also Chrome on Android. So i suggested to remove the appendix and put in the link you also tried (without the `,/Order#banglejs2`).

As for the price, i guess i didnt understand the `£64,£49` then. Whats the 49? But if thats intended, ill leave it in for sure.

gfwilliams:
Ahh - what you're missing is the bit right at the top of the page:

> :warning: **Please view the correctly rendered version of this page at https://www.espruino.com/Bangle.js2. Links, lists, videos, search, and other features will not work correctly when viewed on GitHub** :warning:

The pages are in GitHub so anyone can contribute changes if they want to, but there are specific bits of the markdown that have special meanings when they are transformed for espruino.com - info on those is at https://github.com/espruino/EspruinoDocs/blob/master/README.md - in retrospect maybe they should have been inside Markdown comments or something so they were hidden, but all the docs are made like that now...

So I think this is all working as intended - and you'll actually get a lot more out of the page if you view the transformed version at https://www.espruino.com/Bangle.js2 :)

Out of interest, how did you get to the Github version? Did a search take you there?

PR #684, opened 2022-11-29
Adds support for the MAX17048 and MAX17049 fuel gauges

gniezen:
Adds support for the MAX1704x fuel gauges, as used in the [Adafruit MAX17048 LiPoly / LiIon Fuel Gauge and Battery Monitor](https://www.adafruit.com/product/5580).


gfwilliams:
That looks great - thanks!

PR #686, opened 2022-12-15
Update BLE compatibility of the ESP32, remove NBSP, fix table formatting

Hypfer:
This PR documents the BLE compatibility of the ESP32 which has been available since 2018.
See: https://github.com/espruino/Espruino/issues/1208#issuecomment-388790168

Why in four years apparently none of the ESP32 Espruino users bothered to update the docs is honestly beyond me

Apart from that, I've also replaced all non-breaking spaces with regular ones in the `Other Boards.md` as they seemed to be there by accident and not on purpose.
And finally, all encountered tables have been reformatted

### DIFF
--- a/boards/ESP32.md
+++ b/boards/ESP32.md
@@ -4,7 +4,7 @@ Espruino on ESP32
 
 <span style="color:red">:warning: **Please view the correctly rendered version of this page at https://www.espruino.com/ESP32. Links, lists, videos, search, and other features will not work correctly when viewed on GitHub** :warning:</span>
 
-* KEYWORDS: ESP32,Espruino,Board,PCB,Pinout,Internet,WiFi,Wireless,Radio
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks, that's great! 

> Why in four years apparently none of the ESP32 Espruino users bothered to update the docs is honestly beyond me

Yes, I think it's a general problem with Open Source - for folks doing this in their free time, writing code is more fun than documentation :)

But to be fair, Bluetooth LE wasn't too usable before - I remember having a quick play a year or so ago and seeing a bunch of warnings and not having much success, and while the vast majority of the work had been done on it, it took me a while to fix some niggles (and there are still quite a few outstanding!).

MaBecker:
@Hypfer you are welcome to update the ESP32 docs.

https://github.com/espruino/Espruino/issues/1777

PR #687, opened 2023-01-05
[Tutorial App Customizer] app.json -> metadata.json

thyttan:


gfwilliams:
Thanks - but please could you leave 'Customiser' in the keywords section? `Customizer` is already in there and generally in England we use `ise` rather than `ize`, so potentially someone may search for `Customiser`

thyttan:
Ok, sorry about that! Done!

gfwilliams:
Thanks! In future, there's no worries about changing the title/etc especially as it matches the app loader better - it's just if someone searches for the old thing it'd be better if they can still find it :)

PR #689, opened 2023-02-23
Update Puck.js.md

kurtgrung:
[Getting Started Guide] - link broken

### DIFF
--- a/boards/Puck.js.md
+++ b/boards/Puck.js.md
@@ -128,7 +128,7 @@ Saved code will not be loaded from flash, *but will not be erased from flash eit
 Tutorials
 --------
 
-First, it's best to check out the [Getting Started Guide](/Quick+Start+BLE#puckjs)
+First, it's best to check out the [Getting Started Guide](https://www.puck-js.com/go)
 
 Tutorials using Puck.js:
 


gfwilliams:
Hi - I just wanted to check - how exactly is this link broken?

On https://www.espruino.com/Puck.js#tutorials (which is what this page is used to produce) the 'Getting Started Guide' link appears to work fine and take me to https://www.espruino.com/Quick+Start+BLE#puckjs ?

kurtgrung:
gives a GitHub 404 when i click the link.

gfwilliams:
Where? On https://github.com/espruino/EspruinoDocs/blob/master/boards/Puck.js.md ?

Because we do have right at the top of that page:

> Please view the correctly rendered version of this page at https://www.espruino.com/Puck.js. Links, lists, videos, search, and other features will not work correctly when viewed on GitHub

I have the repo on GitHub in markdown so that it's easy to edit and contribute too, but not everything works when viewed from GitHub

kurtgrung:
yes on > https://github.com/espruino/EspruinoDocs/blob/master/boards/Puck.js.md

gfwilliams:
Ok, thanks - but I'm afraid I don't think this in particular is worth fixing.

I guess maybe at some point someone could come up with a script to change *every* relative link in the repository, so at least those work - but for instance things like `APPEND_USES: ` that you see in GitHub will never work. It's why we have the comment with the big warning sign at the top of every page.

PR #698, opened 2023-07-16
Add examples for GB nav and gps message

halemmerich:


gfwilliams:
Thanks! Not sure how I missed that!

PR #699, opened 2023-08-05
fix command in tutorial

silasms:
Command apparently wrong in 8266 flash download tutorial for windows.


gfwilliams:
Thanks for the update! I'm just looking at what changed though (https://github.com/espruino/EspruinoDocs/pull/699/files) and I don't see any actual difference? It's just that the write_flash command was moved to a new line.

Did the old one not work for you? They should be identical

silasms:
The command was giving me an error, and I organized it that way and it worked! But now it seems to be working, probably I made some mistake. Sorry!

PR #701, opened 2023-08-21
fix waterproofing claim of banglejs1 and save some watches

lorenzo-dev1:
Fix waterproofing claim for banglejs1, following forum discussion: https://forum.espruino.com/conversations/389438/ and destroying my banglejs1 in a swim

### DIFF
--- a/boards/Bangle.js.md
+++ b/boards/Bangle.js.md
@@ -26,7 +26,7 @@ Contents
 Features
 --------
 
-* IP68 Waterproof: up to 10m underwater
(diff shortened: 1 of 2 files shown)


gfwilliams:
Argh, I'm really sorry to hear that - thanks for the documentation update.

The 10m claim was what the original watch manufacturer did say the watch was capable of, but it has seemed that in many cases they haven't been capable of anything near that, so for the Bangle.js 2 where we had similar problems we'd changed the wording - unfortunately we hadn't for the Bangle.js 1

PR #706, opened 2023-12-18
buildmodules.sh in now javascript script build file

brendena:
I moved the build **buildmodules.sh** script into javascript file **./bin/buildmodules.js** .

command now
- node run build_modules 

Goal
- Add the ability to build just one file via **--file=./device/name**
- Simplify the build by moving more to node build cycle
- Potentially make it run faster


brendena:
Not entirely sure if this is a good change.  But i made it.

gfwilliams:
Has `minify.js` just turned into a module so it can no longer be used from the command-line?

Because we use it from the Espruino repo, for instance https://github.com/espruino/Espruino/blob/master/libs/js/update_js.sh

I use `buildmodules.sh` from scripts here to build the main site as well, and even `build.sh` in this repo uses it and hasn't been updated so would break.

Thanks for this, but I'm not sure I want to pull it in I'm afraid - if you wanted to just extend `buildmodules.sh` to optionally handle one file I could understand, or if this added extra features, but just swapping to JS instead without really adding functionality, and in the process breaking EspruinoDocs, Espruino, and my internal tools just seems a bit too much of a stretch...

Unfortunately I'm not sure it's going to speed stuff up either - most of the heavy lifting is done in Java which would be reloeaded for every file anyway.

PR #707, opened 2023-12-18
universal_presentation_clicker - wrong image in the page

AkosLukacs:
Incorrect image shown in the page https://www.espruino.com/universal_presentation_clicker


gfwilliams:
Thanks! Good spot!

PR #708, opened 2023-12-21
E-ink driver for SSD16__ devices

brendena:
Its currently only have it working with the displays i have.  But it seems that many of these controllers have a similar structure.  So my hope is for people to try it with there display and make some small changes if needed.


gfwilliams:
Wow, this looks excellent - thanks!

PR #714, opened 2024-01-15
fixed a cors problem when serving files locally

brendena:
So this does two things.  The module files don't get served with the server because there built in a different directory.  

Then there seems to be a CORS problem.  I'm guessing because the ide has a origin of **https://www.espruino.com** and the express server just rejects it.  So i just set it to allow anyone to receive the file.  

Fixes this error
Access to fetch at '[http://localhost:3040/modules/BH1792.min­.js'](http://microcosm.app/out/Zb5Si) from
origin '[https://www.espruino.com'](http://microcosm.app/out/5c5Si) has been blocked by CORS policy: No
'Access-Control-Allow-Origin' header is present on the requested
resource. If an opaque response serves your needs, set the request's
mode to 'no-cors' to fetch the resource with CORS disabled.

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,19 +1,26 @@
 /* node.js script that creates a webserver to serve up the pages from the html directory */
 
 var express = require('express');
+const cors = require('cors');
+const homeDirectory = require('os').homedir();
 var app = express();
+app.use(cors());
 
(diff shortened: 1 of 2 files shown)


gfwilliams:
This looks great, thanks!

PR #715, opened 2024-01-18
Remove `format` from `showMenu`

bobrippling:
This is handled automatically by `showMenu` and now will trigger a sanity check.

See espruino/BangleApps#3149


gfwilliams:
Great - thank you!

PR #716, opened 2024-01-20
Cleaned up the docs a bit and added more privacy friendly options

Darin755:
This pull request rewords a few things and adds Ungoogled Chromium, Chromite and the Android Banglejs app. The docs still need a lot of work but this will add a few privacy friendly and FOSS options.

### DIFF
--- a/tutorials/Quick Start BLE.md
+++ b/tutorials/Quick Start BLE.md
@@ -145,7 +145,7 @@ Using the Espruino IDE
 
 If your computer supports it, Web Bluetooth is the easiest way to get started with Espruino.
 
-You'll need an up to date chromium browser such as [Google Chrome](https://www.google.com/chrome/browser/desktop/), [Ungoogled Chromium](https://ungoogled-software.github.io/ungoogled-chromium-binaries), [Edge](https://www.microsoft.com/en-us/edge) or [Opera](https://www.opera.com) (Brave will not work). Operating System support varies so for platform specific instructions see below:
+You'll need an up to date Chromium browser such as [Google Chrome](https://www.google.com/chrome/browser/desktop/), [Ungoogled Chromium](https://ungoogled-software.github.io/ungoogled-chromium-binaries), [Edge](https://www.microsoft.com/en-us/edge) or [Opera](https://www.opera.com) (Brave will not work). Operating System support varies so for platform specific instructions see below:
 
 
 #### Mac OS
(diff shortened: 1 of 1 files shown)


tutorials/Quick Start BLE.md
```diff
@@ -98,9 +98,8 @@ Requirements
 For Bluetooth LE you need a Bluetooth 4.0-capable adaptor in your computer (Bluetooth versions before 4.0 won't work). Pretty much all new computers come with Bluetooth 4, but you *may* need to get an external Bluetooth LE dongle if your computer:
 
 * Is an Apple Mac made before 2012
-* Is a Windows PC with a Windows version before 10
+* Is an older Windows or Linux device that predates Bluetooth 4.0
```
gfwilliams, line 101: I'm not sure this is a helpful change? How are people supposed to know if their device predates Bluetooth 4? Saying before windows 10 was accurate and easy to check.

Darin755: You can run Windows 10 on a device with Bluetooth 3.0

Perhaps we could add both? How about something like "a device older than Windows 10 or a device that doesn't support Bluetooth 4.0)

Darin755:
> Thanks! Is it possible you could fix the capitalisation on the stuff you'd added? Things like `chromium browser` and `bangle.js`?
> 
> Also when you mention the Bangle.js app you should probably say Gadgetbridge and link to https://www.espruino.com/Gadgetbridge for the install instructions so we're not duplicating links. It also doesn't provide the Web IDE so I'm not sure if users wanting to develop would find that helpful?

Sorry about the capitalization, I must of missed it. I included the banglejs app because it has better support out of the box for Banglejs. Gadget bridge doesn't provide a way to load apps or update Agps data for example.

gfwilliams:
Thanks! 

>  included the banglejs app because it has better support out of the box for Banglejs. Gadget bridge doesn't provide a way to load apps or update Agps data for example.

Ahh, ok - there's a bit of miscommunication there - the Bangle.js app *is* Gadgetbridge, it's just a version with internet access allowed. That page I linked explains the situation and all the places you can get it, so really that's the one to link to.

Darin755:
That does make sense. I'll go ahead and link to the separate page.

Darin755:
I have updated my fork with the proposed changes. Just let me know if you need anything else that needs updated.

PR #718, opened 2024-03-11
android:info re activity fetching of recorder logs

thyttan:


gfwilliams:
Thanks! Do we need to mention what gets sent from the Bangle to the phone? There's `actTrk` and `actTrksList`?

thyttan:
I added the info - not super happy with how I structured it but good enough maybe?

gfwilliams:
Looks good to me - thanks!

It's really just about having something down so that folks have somewhere to start looking

PR #719, opened 2024-03-11
Update clock_info docs with focus/blur

bobrippling:
This updates the docs to match `clock_info` changes from espruino/BangleApps#3234


gfwilliams:
Thanks!

PR #720, opened 2024-03-12
Fix broken link to "Web IDE" on Firmware Update page

CarpeNecopinum:
Clicking the old link on espruino.com leads to `https://www.espruino.com/espruino.com/ide`, which results in a 404 error.

(A few paragraphs up, there was a working link, I just copied that over, thus the added `www.`)


gfwilliams:
Thanks, that's great! It's really nice to catch these kinds of problems

PR #721, opened 2024-03-20
Deprecate currency information in `locale`

atjn:
Closes https://github.com/espruino/BangleApps/issues/3269


gfwilliams:
Thanks!

PR #725, opened 2024-04-02
New tutorial: Testing Espruino on Bangle.js 2

### LINKED ISSUE #723: New tutorial: How to rapidly build and test Espruino for banglejs2
This is a continuation of https://github.com/espruino/Espruino/pull/2476#issuecomment-2009488313

I want to quickly change some code in Espruino, then compile it and flash it to my watch or the web ide emulator so that I can test my changes. That process should be very fast and very safe - no risk of breaking my watch just because I want to test a small change.

I am not able to find any documentation that can help me with that, so I want to write a tutorial specifically for this use case. I will probably need your help in finding a solution that works at all.

It should cover:

- [x] How to compile espruino - you mentioned the github action, that seems like the best solution
- [x] How to upload it easily to the banglejs2. Preferrably you can do this directly from the PC you are developing on via bluetooth.
  - [x] Using the App Loader's `fwupdate`
  - [x] Need to unzip first - suggested solution in https://github.com/espruino/BangleApps/pull/3298
- [x] How to make sure the upload cannot brick the device
- [x] Alternatively, how to build the Web IDE with an emulator that uses the custom espruino build
  - [x] How to compile the emulators - suggested solution in https://github.com/espruino/Espruino/pull/2479
  - [x] How to easily run a local copy of the Web IDE - suggested solution in https://github.com/espruino/EspruinoWebIDE/pull/295

I think this would be a very useful guide, what do you think?

atjn:
Closes #723


gfwilliams:
This looks great - thanks!

PR #728, opened 2024-04-12
add new modue hmac256

MaBecker:
based on hmac


gfwilliams:
Thanks! But wasn't it possible just to add this single function to HMAC?

```
exports.SHA256 = function(key) {
  return new exports.HMAC(key, require('crypto').SHA256, 64, 32);
};
```

Or was there something else?

MaBecker:
Wow, ok looks like have been way much over motivated..... Sure why not just, one file. What happens if `require('crypto').SHA256` failes because sha256 is not available?

gfwilliams:
Sorry! Well, it'll only fail when the function is executed - there's no problem with it being in there and SHA256 not being supported if it's not used.

MaBecker:
please check again.

gfwilliams:
Looks great - thanks for doing that!

PR #729, opened 2024-04-16
Update BTHome.js

yerpj:
Support for CO2 data

PR #730, opened 2024-04-17
Update BTHome.md

yerpj:
update BTHome documentation
To comply with new version of BTHome module. See [here](https://github.com/espruino/EspruinoDocs/pull/729) for details


gfwilliams:
Thanks!

PR #731, opened 2024-04-20
allow load from storage

MaBecker:


gfwilliams:
Thanks! What do you mean by "allow load from storage" though?

Is it literally just that you're trying to save by putting the module in quotes, and then the escape codes would need double quoting? If so we can't go around removing all escape characters from every module.

What if you did like I suggested in the forum? Literally just using the 'Upload' button built into the IDE's Storage menu?

MaBecker:
> Thanks! What do you mean by "allow load from storage" though?
> 
> Is it literally just that you're trying to save by putting the module in quotes, and then the escape codes would need double quoting? If so we can't go around removing all escape characters from every module.
> 
> What if you did like I suggested in the forum? Literally just using the 'Upload' button built into the IDE's Storage menu?

Ok, will give it a try. Couldn’t mage to upload the last time.

PR #734, opened 2024-06-15
Renesas FS3000 air velocity sensor

AkosLukacs:
Driver for Renesas FS3000.
Tested on Espruino Wifi, but it's so basic it should run on anything.


gfwilliams:
This looks great - thank you!

PR #735, opened 2024-08-27
Document use of meridian

atjn:
Implements solution 3 in https://github.com/espruino/BangleApps/issues/3534


gfwilliams:
Looks great - thanks!

PR #736, opened 2024-09-11
Use `locale` in clock tutorial

atjn:
This rewrites the clock face tutorial to use the `locale` module instead of writing its own basic clock generator.

The new code also uses meridians correctly, which implements solution 4 in https://github.com/espruino/BangleApps/issues/3534


gfwilliams:
Thanks!

gfwilliams:
If you see any other stuff this (updating tutorials to suggest the newer/better way of doing things) it'd be great if you could do PRs, or even just bring it to my attention!

PR #738, opened 2024-09-29
Remove duplicate meridian

ruvilonix:
Meridian is already included in time(), so calling meridian() makes the string have two meridians.

### DIFF
--- a/tutorials/Bangle.js Clock.md
+++ b/tutorials/Bangle.js Clock.md
@@ -23,9 +23,7 @@ Copy the following code to the right of the IDE and click Upload ![](data:image/
 function draw() {
   // work out how to display the current time
   var d = new Date();
-  var clock = require("locale").time(d);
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! Sorry about this.

It looks like the 'default' locale contains meridian (https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/locale.js) but none of the other locales you might install with the Languages app do. @atjn 

I've just pushed a change to the main Bangle.js firmware so that on a cutting edge firmware (or 2v25 when released) the meridian will be removed from the built-in locale so it matches what happens elsewhere.

So for now if developing with a clock I'd do what we mention in the tutorial and just install the `Languages` app (and soon the 2v25 firmware will come out and will fix this by default).

atjn:
I must have made a mistake when testing this. Sorry about that :)

PR #740, opened 2024-11-16
Make font converter keyboard compatible

KTibow:
The input currently requires your mouse to move over it for it to update. This is (very) annoying if you're changing it with your keyboard. This PR makes the update trigger on any `input`.


gfwilliams:
Great - thanks!

PR #745, opened 2025-03-20
Add library for decoding the Quite Ok Audio Format (QOA)

ssievert42:
This adds a library for decoding the [Quite Ok Audio Format (QOA)](https://qoaformat.org/), using some inline C and a fancy js wrapper.

QOA does reasonably fast lossy audio compression from 16 bits per sample down to 3.2 bits per sample, and should allow for saving some precious flash.

This PR kind of supersedes https://github.com/espruino/Espruino/pull/2592, since it pretty much does everything that PR did and then some more. (And only uses additional flash if a user wants to decode QOA encoded audio, instead of having to be baked into the firmware, not being used by most and wasting space.)

The docs for the library include a QOA encoder in pure js to make it easier for people to encode their own audio files to QOA. (Thanks for the starter code by the way, that helped a lot! https://github.com/espruino/Espruino/pull/2592#issuecomment-2592105727)
Alternatively, the [reference encoder / decoder](https://github.com/phoboslab/qoa) can be used as well. (The docs for the library describe how to use `ffmpeg` and the reference encoder to produce some 8 kHz mono audio.)

I tested and developed this using Jolt.js with 8 kHz mono audio.

Basic usage looks like this:
```js
let qoa = require("QOA");
qoa.play("audio.qoa", {output: "waveform", outputOptions: {pin: H0, pin_neg: H1}});
```
But for anyone wanting to do the playback themselves, decoding samples to a buffer is exposed as well.

Some (fairly loopable) zipped example files to quickly test playback (zipped because GitHub doesn't allow uploading files ending in `.qoa`):
- [radio.zip](https://github.com/user-attachments/files/19375053/radio.zip)
- [moveIt.zip](https://github.com/user-attachments/files/19375070/moveIt.zip) (if you feel like annoying yourself or someone else)

Weirdly enough, playback is slow and seems choppy when connected via BLE, but is a normal amount of speedy and doesn't seem to be choppy when connected via serial.
I though playback speed was related to the voltage the board is powered at, but the difference when connecting via serial and disconnecting is quite enourmous.

Eventually, I'd like to play audio over I2S for that extra quality boost, but I always seem to run out of time when thinking about hacking I2S support into Espruino. (Which is the reason for adding the `output` option to the `play()` function; to be able to potentially support I2S in the future.)

I'd love to get some feedback on this :)

### DIFF
--- a/modules/QOA.js
+++ b/modules/QOA.js
@@ -1,0 +1,599 @@
+/*
+Copyright (c) 2025 Simon Sievert
+SPDX-License-Identifier: MIT
+*/
+
+/*
+Library for working with the [Quite Ok Audio Format (QOA)](https://qoaformat.org).
+*/
+
+let qoa = E.compiledC(`
+//int SAMPLES_PER_FRAME()
+//int ENCODED_FRAME_SIZE_BYTES()
+//int MIN_FILESIZE()
+//int DECODER_STATE_SIZE()
+//void set_decoder_state_buf(int)
+//void set_use_16_bit(bool)
+//bool get_use_16_bit()
+//int get_last_frame_len()
+//void fill_with_silence(int, int)
+//int decode_header(int, int)
+//int get_sample_rate()
+//int get_total_samples()
+//int decode_frame(int, int, int)
+//void copy_data(int, int, int)
+//void clear_buf(int, int)
+
+typedef signed char int8_t;
+typedef unsigned char uint8_t;
+typedef signed short int16_t;
+typedef unsigned short uint16_t;
+typedef signed int int32_t;
+typedef unsigned int uint32_t;
+
+typedef unsigned int size_t;
+
+/*
+ * Copyright (c) 2023, Dominic Szablewski - https://phoboslab.org
+ * SPDX-License-Identifier: MIT
+ *
+ * QOA - The "Quite OK Audio" format for fast, lossy audio compression
+ */
+
+
+static const uint32_t QOA_MIN_FILESIZE = 16;
+static const uint32_t QOA_MAX_CHANNELS = 8;
+
+static const uint32_t QOA_SLICE_LEN = 20;
+static const uint32_t QOA_SLICES_PER_FRAME = 256;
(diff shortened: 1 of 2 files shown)


gfwilliams:
This looks absolutely awesome - thank you! It's so nice to have the encoder built into the reference pages!

This will be really useful for a while bunch of people.

Very interesting about Bluetooth having such an effect too - does `NRF.setConnectionInterval(200)` make things better (albeit at the cost of a laggy/slower REPL and upload)? By default Espruino goes into high speed mode at 7.5ms and maybe that ends up using a bunch of CPU time.

It's possible that it's not so much a timing issue, as it just being that with BLE active the bluetooth stack is using enough CPU time that there's not enough left for audio. I believe it should be possible to do audio using DMA which would obviously improve matters greatly, but the Waveform class does have the benefit that it's polyphonic (so you can play more than one waveform at the same time, even at different speeds, and they're played back correctly).

ssievert42:
Thanks :)

> does NRF.setConnectionInterval(200) make things better

That seems to help a little :+1:
But not nearly as much as connecting via serial.
So in short:
- not connected at all: slow and stuttery audio
- connected via BLE: slow and stuttery audio
- connected via BLE and doing `NRF.setConnectionInterval(200)`: slightly less slow and stuttery audio
- connected via BLE and doing `NRF.sleep()`: slightly less slow and stuttery audio (now this, I don't understand)
- connected via USB/serial: perfect audio

I'd have assumed that not being connected at all should be pretty similar to being connected via serial, but yeah, looks like the BLE stack is eating all the CPU cycles. (Even when it probably shouldn't?)

> you can play more than one waveform at the same time

Neat! That's pretty much exactly what I was gonna do next: Try to do some audio mixing to play music and effects at the same time; The `play()` function could probably use an option to tell it to not mess with the pins for that exact use case.

ssievert42:
Seems like minifying with the closure compiler kind of wrecks the inline C, among other things replacing backticks with regular quotes and, most importantly, causing the Espruino tools to not replace `E.compiledC()` calls :see_no_evil:

gfwilliams:
Oh wow, thanks. That's a pain - I've just looked into it and put a fix in if you reload the IDE now - your QOA module should be live too

PR #749, opened 2025-05-16
Gadgetbridge: Add http insecure flag

joserebelo:
Documents the change from https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/4845


gfwilliams:
Great - thanks!

PR #752, opened 2025-09-10
A module for Sensiron SCD40

Radyr:
My first attempt at a module for the Sensiron SCD40.

It should also work with the SCD41 and SCD43 but not tested as I dont have those to hand.


gfwilliams:
This looks great, thanks! Please could you convert:

```
  REQUEST_SINGLE_LOOKUP : Uint8Array([0x219d>>8,0x219d]),
  REQUEST_SINGLE_LOOKUP_RHT : Uint8Array([0x2196>>8,0x2196]),
  // and so on
```

to just:

```
  REQUEST_SINGLE_LOOKUP : [0x21,0x9d]),
  REQUEST_SINGLE_LOOKUP_RHT : [0x21,0x96]),
  // and so on
```

While Uint8Array is good for big arrays of bytes, there's quite a bit of overhead for just storing 2 bytes worth of data, so in this case just storing the data in a standard array should be a lot better.

PR #759, opened 2026-06-14
gadgetbridge.md: tweak docs re intent extras

thyttan:
To indicate extras can be whatever value. Although not arrays as of now per https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/6201 .

### DIFF
--- a/info/Gadgetbridge.md
+++ b/info/Gadgetbridge.md
@@ -144,10 +144,10 @@ The following type of information can be supplied for intents: `target`, `action
 Template for initiating an intent from a Bangle.js app:
 
 ```
-Bluetooth.println(JSON.stringify({t:"intent", target:"", action:"", flags:["flag1", "flag2",...], categories:["category1","category2",...], package:"", class:"", mimetype:"", data:"", extra:{someKey:someValue, anotherKey:"someString",...}}));
+Bluetooth.println(JSON.stringify({t:"intent", target:"", action:"", flags:["flag1", "flag2",...], categories:["category1","category2",...], package:"", class:"", mimetype:"", data:"", extra:{someKey:valueType, anotherKey:valueType,...}}));
 ```
 
-Key/value-pairs can be omitted if they are not needed.
+Key/value-pairs can be omitted if they are not needed. `valueType` can be a boolean (`true/false`), integer, floating point, or String - and each value in `extra` can be a different type.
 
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! I just added some extra text to try and clarify things a bit more too.

thyttan:
It was further extended now, here: https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/6284

I didn't understand if an extra can now also be a JSONObject (w key/value pairs) or if it's mainly just arrays in that PR.

Edit: seems like passing objects via extras are usually done by turning them into strings first. https://stackoverflow.com/questions/5082122/passing-jsonobject-into-another-activity

So seems like extras can't contain bare JSONObjects (w key/value pairs) anyways.

gfwilliams:
Ok, thanks! So likely it's really just arrays of strings (as Jose tested) that make sense to mention?

thyttan:
Arrays of booleans, doubles, longs and strings it seems from ThomasKuehnes description and Josés test:


>        "booleanArray": [true, false],
>        "doubleArray": [1.2, -3.4],
>        "longArray": [12, -34],
>        "longArray2": [2147483649],
>        "StringArray": ["12", "xyz"]

thyttan:
Seems weird there should not be usual integer arrays no?

gfwilliams:
It does, but I think it's just he forgot to mention it. There is code for it?

```
if (value instanceof final int[] extra) {
            bundle.putIntArray(key, extra);
```

ThomasKuehne:
JSON only supports integers without making a distinction between `int` and `long`. The JSON array decides for each individual value if to use Integer or Long - so you could end up with {int long, int ...}.

So JSON arrays starting with an int or a long are always handed over as `long[]` to the Intent:
https://codeberg.org/Freeyourgadget/Gadgetbridge/src/commit/8943b33597e199bb371605e5fcbadede5db81126/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/BundleUtils.java#L231-L232

JSON structures without special handling are handed over as Strings to the Intent:
https://codeberg.org/Freeyourgadget/Gadgetbridge/src/commit/8943b33597e199bb371605e5fcbadede5db81126/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/banglejs/BangleJSDeviceSupport.java#L1074-L1075

PR #760, opened 2026-06-14
gadgetbridge.md: Add device extra to global tx intent

joserebelo:
As of https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/6290


gfwilliams:
Great - thanks!

